Issue #17, opened 2020-07-08
Rewrite Python Test Framework docs


kazk:
* [x] reference
* [ ] guide/tutorial

---

We currently have two Python test framework docs.

- https://github.com/codewars/docs/blob/master/wiki/codewars-python-test-framework.md
- https://github.com/codewars/docs/blob/master/wiki/codewars-python-test-framework-v2.md

Based on these, we should create:

- A guide for writing good tests with it (read few times when starting)
- A reference manual you can refer to (read multiple times)

The guide should describe how to write tests. The reference should list all the assertion functions and example usages.

Later, we can also add another page for specific topics, e.g., writing random tests.

Feel free to comment for any thoughts. Also, assign yourself to let the other members know you're going to work on it.

Blind4Basics:
do we agree to describe only v2? (sying that v1 still exists in old katas should be enough. Only PUs will be facing old codes to maintain anyway, I guess)

I'll begin with the reference manual. How do I proceed? pull request on doc.codewars, I guess?

Blind4Basics:
BTW, now that we're talking about it... Should we add my module forbidder to the test framework? That would require some work, but it's already in the wild in a lot of katas and it's spreading. And it has some flaws and it already became impossible to update it everywhere, since I don't know all the places... :/

(yeah, it's "a bit" off topic... x) )

kazk:
Yeah, focus on the v2. We can add some notes about the old one later.

For module forbidder, can you open a new issue at https://github.com/codewars/python-test-framework?

Blind4Basics:
done

is that repo the last version? (I believe not?) I'd like to have the code on hand so that I can check what I write about it. XD

kazk:
Yeah, it's the version being used on Codewars.

Blind4Basics:
:+1:

Blind4Basics:
how do I integrate an picture...? :/

Blind4Basics:
https://github.com/codewars/docs/pull/18

kazk:
Can you move the files?

- document to `content/languages/python/codewars-test-famework/index.md`
- image to `content/languages/python/codewars-test-framework/python-test-framework-example.png`

Then, use the image in the Markdown. `![Python Test Framework Example](./python-test-framework-example.png)`.

Also, add frontmatter to Markdown so that it shows up on the Python page and navigations are added.

```
---
category: references
languages: [python]
tags:
  - testing
prev: /languages/python/
---
```

[...]

PR #18, opened 2020-07-08
Add reference for Python Test Framework

Blind4Basics:
Reference documentation

* I removed the example at the end (it will go in the guide)
* missing the information about the module forbidder, if it's finally incorporated
* missing the image (uploaded, but didn't find how to insert it in the md file) 
  -> maybe putting it into `static`? then how do iI access it? `` ?
* needs an extra check about the english... XD

* I forgot the headers (tags/categories, ...)


kazk:
By the way, to see how the document will appear, click "Details":

image

and go to the path of the document you're editing.

Blind4Basics:
since I forgot to put the headers, I guess it won't be visible at all for now...? XD

kazk:
It's visible, but hard to find. I added the URL to my comment above.

Blind4Basics:
yep, I saw, thx. :)

ok, looks nice enough. I messed up the Overniwe markdown (some sections do not appear in the right sidebar, for now

content/codewars-python-test-framework-v2.md
```diff
@@ -0,0 +1,298 @@
...
+
+<br>
+<br>
+
+
+
+---
+
+
+# Tests interruption
```
kazk, line 100: Only use `h1` (`# heading`) once for the title of the document.

Blind4Basics: yep, I made the correction.

content/codewars-python-test-framework-v2.md
```diff
@@ -0,0 +1,298 @@
...
+
+The test framework allowes to write named groups of tests holding other named groups of tests, containing themselves assertions.
+
+* Assertions (using the methods detailed below) can actually be used at any level in the groups, or even outise of them (but you're advised to avoid that).
+* Note that the code executed inside on group isn't stopped at the first failed assertion (unlees you use `allow_raise`. See below), so the full test suite is executed each time.
+
+The basic setup for the tests follow this example:
+
+```python
+@Test.describe('Fixed Tests')
```
kazk, line 26: Can you change to `@test`?

We'll continue to support `Test` for now, but in the future, I'd like to let users just import the test framework normally like `import codewars_test as test` (removing the runner magic). So I want the documentation to constantly use `test`.

content/codewars-python-test-framework-v2.md
kazk: ? I meant once at the top `# Python test framework for Codewars (v2)`. There are still this one and `# Available assertion methods`. Those should be `##` and any children increased by one level.

content/codewars-python-test-framework-v2.md
Blind4Basics: argh... I just did the complete opposite x)

Ok.

content/codewars-python-test-framework-v2.md
Blind4Basics: mmh? I thought I already corrected that one too. I'll check

content/codewars-python-test-framework-v2.md
Blind4Basics: done (the markdowns too)

content/codewars-python-test-framework-v2.md
```diff
@@ -0,0 +1,298 @@
...
+
+---
+
+
+## Tests interruption
+
+Some of the functions below can accept a named argument `allow_raise=False`.
+
+If you change its value to `True`, the tests contained inside the current block will be interrupted at the first failed test. The executions are then going back to the parent block if it exists and the next part is executed.
+On some computation-wise intensive katas, this may be a good idea to use this feature so that the user doesn't have to wait a long time before getting feedback (or possibly befor etiming out and in that cas, he might never get ny feedback at all, which may be cumbersome).
```
kazk, line 105:
```suggestion
On some computation-wise intensive katas, this may be a good idea to use this feature so that the user doesn't have to wait a long time before getting feedback (or possibly before timing out and in that case, they might never get any feedback at all, which may be cumbersome).
```

Blind4Basics:
* all required modifications done (as well as wath's in https://github.com/codewars/docs/issues/17#issuecomment-655784221
* reread the full doc completely and made (a lot... XD ) of corrections

kazk:
Thanks, I think it's good enough for a first draft. We can iterate on this.
I'll make some changes after merging.

PR #28, opened 2020-07-09
Clean up Python test framework doc

kazk:
- Remove v2 from title
- Move acknowledgement to last
- Reword headings
- Don't mention assertions _can_ be anywhere
- Clarify `Test` alias
- Remove unnecessary `<br>` and `<hr>`
- Remove information about the current output. This is about the framework and shouldn't include minor details about the output.

### DIFF
--- a/content/languages/python/codewars-test-famework/index.md
+++ b/content/languages/python/codewars-test-famework/index.md
@@ -7,28 +7,17 @@ sidebar: "language:python"
 prev: /languages/python/
 ---
 
-# Python test framework for Codewars (v2)
+# Python Codewars Test Framework
 
-(upgrade by [@Bubbler-4](https://github.com/Bubbler-4), Feb 2019)
+Codewars currently uses a [custom test framework][test-framework-repo] to test Python.
 
-<br>
-<br>
+The framework is automatically imported and assigned to `test`.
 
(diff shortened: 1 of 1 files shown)


Blind4Basics:
> Remove unnecessary <br> and <hr>

x)  
Well, I find those usefull when ones want to go quickly though the doc (visually). Is there a specific reason to not give spaces around the different sections?

kazk:
Consistency. We should change the CSS if more vertical space improves readability or maybe increasing the font size of the headings will help. Adding arbitrary `<hr>` and `<br>` is not a good idea.

kazk:
For `<hr>` above the heading, I don't see any improvement because the heading clearly starts a new section.

With `<br><br><hr>`:

image

without:

image

`<hr>` should be used when separating siblings (in terms of document hierarchy, not DOM tree) and only when it makes sense.

kazk:
If you haven't already, please check how it looks on https://docs.codewars.com. To me, removing `<br><br>` doesn't make significant difference and removing `<hr>` clarifies the hierarchy. These are why I wrote unnecessary.

If you find this change actually made it less readable, I'd like to know so that we can change the overall style instead.

Blind4Basics:
well, it's rather a matter of taste than anything, I guess.

Consistency/CSS, right. :+1:

I'd personnaly like to have more space above the different kind of titles.  
The `hr` line is totally forgetable, yes.

kazk:
Yeah, we can increase the spaces above headings a little. Also, it's probably better to adjust the space by heading level. So for example, more space above `h2` than above `h4`.

PR #36, opened 2020-07-13
Update glossary.yml

Blind4Basics:
do not merge yet: I'm making tries about internal links from one point to another of the glossary

### DIFF
--- a/data/glossary.yml
+++ b/data/glossary.yml
@@ -73,7 +73,7 @@
 
 - term: Console
   description: Synonym for "output panel" (see below).
-  
+
 - term: Contributor(s)
   description: User(s) who have either authored, approved, translated or modified a kata in any way (a list of them can be seen in the "Details" page of any kata.
 
--- a/data/glossary.yml
+++ b/data/glossary.yml
@@ -39,7 +39,7 @@
 
 - term: Allies
   description: Users who are following each others, or who are part of the same clan.
-  
+
 - term: Approve (a kata, a translation)
   description: |
     Action to push a kata out of the beta process, giving it a rank (kyus). This can be done by power users.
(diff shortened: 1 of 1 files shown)


Blind4Basics:
DX

How the f... can I get useful feedback about what I'm doing wrong, in there? x/

(where do I drop the usual `Don't use Test.expect` thing? x// )

kazk:
Just check the output:
image

image

kazk:
You're using the colon again in unquoted string.

Blind4Basics:
Damn it... XD

thx

kazk:
Maybe just always use the multiline string.

Blind4Basics:
Ready for reviewing.

Note: how is it that sometimes the lint check just hangs forever? (seems like it happens when the "Prettier" thing is showing up)

kazk:
> Note: how is it that sometimes the lint check just hangs forever? (seems like it happens when the "Prettier" thing is showing up)

Yeah, that's auto committed by a tool when there's an issue with formatting and apparently it doesn't re-trigger the checks. I'm looking into an alternative way.

data/glossary.yml
```diff
@@ -138,6 +167,47 @@
...
+    - title: Top 500 by completed Katas
+      url: https://www.codewars.com/users/leaderboard/kata
+    - title: Top 500 by authored Katas and translations
+      url: https://www.codewars.com/users/leaderboard/authored
+    - title: Your personnal leaderboard (with your allies, in the dashboard)
+      url: https://www.codewars.com/dashboard
+
+- term: Markdowns (formatting)
+  description: |
+    CodeWars is responsive to `markdowns`. They allow you to format code, descriptions or comments so that thay are properly readable for other users.
```
kazk, line 188:
```suggestion
    Codewars is responsive to `markdowns`. They allow you to format code, descriptions or comments so that thay are properly readable for other users.
```

kazk:
Thanks. We'll iterate on some descriptions later to clarify things.

Issue #23, opened 2020-07-09
Complete the glossary (an ongoing comprehensive (?) list)


Blind4Basics:
drop here stuff to define.

To keep this usable: copy the whole message each time and insert your suggestions in alphabetic order. 
Once we begin to complete the glossary, we'll remove the words on the fly.

* dashboard
* example tests
* fixed tests
* honor
* kata
* kata (search) page
* kumite
* labels
* markdowns
* mutation of the input
* progress
* random tests
* rank
* sample tests
* spoiler flag
* test suite
* test cases

Question: 

Is it possible to have the definitions linking/referencing themsleves? Something like:

```
example tests: this is, blablabla

...

... 

sample tests: see "example tests"  (here with a link to the definition)
```

hobovsky:
> Is it possible to have the definitions linking/referencing themsleves?

I believe that's how `see`, and possibly `link` sections work in the template described here? https://github.com/codewars/docs/issues/9#issuecomment-655477654

kazk:
`see` is used for internal link. Set this to a path to the document and the title will be automatically filled. It'll show up under the description like `See Title`.

`links` are used for external links. You need to specify the `title` and the `url`.

https://github.com/codewars/docs/blob/f45064a7d07bb5ae50d75d0c423a47233ed5e56d/data/glossary.yml#L28-L32

Blind4Basics:
are some formatting elements possible in the descriptions? (bold, italic, colors, ...?) I guess it's only raw text?  
Multiline descriptions, maybe?

Blind4Basics:
# EDITING

;p

kazk:
Description supports Markdown.
https://github.com/codewars/docs/blob/f45064a7d07bb5ae50d75d0c423a47233ed5e56d/data/glossary.yml#L12-L13
https://github.com/codewars/docs/blob/f45064a7d07bb5ae50d75d0c423a47233ed5e56d/data/glossary.yml#L62-L68

![image](https://user-images.githubusercontent.com/639336/87119526-277f5b80-c233-11ea-9651-6ce7767ec78e.png)

https://docs.codewars.com/glossary/

hobovsky:
Suggestions:
 - contributor
 - beta kata
 - to accept/to approve (whichever is correct term) a beta or a translation
 - to reject
 - retired kata
 - fork
 - merge conflict
 - abandoned kata
 - active/inactive author
 - to forfeit (a kata)
 - trainer suggestions
 - issue
 - question
[...]

PR #47, opened 2020-07-14
Update glossary.yml

Blind4Basics:
Note: I didn't "cross out" from #23 the items added in the previous update so that we can still find/review them more easily.

For this update, changes start at "Random tests"

### DIFF
--- a/data/glossary.yml
+++ b/data/glossary.yml
@@ -108,7 +108,7 @@
 
 - term: Fixed tests
   description: |
-    Tests whose input and output are always the same, each time your run the tests. Generally, the example tests are made of those, and the test cases contains both fixed and random tests.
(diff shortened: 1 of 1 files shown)


Blind4Basics:
It's awfully slow, but looks like the "prettier" thing did work, this time :)

( \o/ woohoo! PR valide at first try! x) )

kazk:
> It's awfully slow

Looks like the checks and auto commit happened within few minutes, so it's not slow (for a CI setup). But the dependency cache wasn't found and the majority of the time was spent installing them, so I'll look into that.

> but looks like the "prettier" thing did work, this time :)

Yeah, I fixed that in #44.

kazk:
I think #48 fixes the caching issue so the next PR should be faster.

Issue #40, opened 2020-07-13
Add Getting Started doc


kazk:
We need some documentation for new users. This document should provide information to help them get started without overwhelming them with too many new concepts.

We should discuss what needs to be included first.

Placeholder: https://github.com/codewars/docs/blob/master/content/getting-started.md

Feel free to comment for any thoughts. Also, assign yourself to let the other members know you're going to work on it.

hobovsky:
I was thinking how to tackle such idea and I came up to a couple of conclusions. I ended up with idea which can be described with following conspect:

 - **Creating account** (is it relevant? Will docs be available for a user who's not signed in?)
 - **Setting up your account**, containing information on:
   - account details,
   - training setup
   - selected languages
 - **Getting a kata to solve**
   - training suggestions,
   - kata search page
   - kata "numbers": rank, satisfaction, possibly others (which?) as criteria for selecting a kata to solve
 - **Solving a kata**
    - Trainer, submitting a solution
    - Solution troubleshooting
    - forfeiting a kata
    - satisfaction voting
 - **Solutions**: viewing, voting,
 - **Participating in discussion**
    - Discourse, posts, spoilers, post labels
    - out-of-site places: gitter, zulip, forum

I think "Getting started" does not need to cover things like translating or authoring a kata, or kumite, because these are not "first contact" activities. I focused on things needed for users being on-boarded.

What do you think?,

kazk:
Looks great.

> Creating account (is it relevant? Will docs be available for a user who's not signed in?)

Anyone can view the docs so it might be worth having a page that can be linked from the getting started page (e.g., "If you don't have an account yet, see ...").

I think having separate pages for each of the areas and using the `prev`/`next` links will be good. We can also have a "getting started" sidebar listing all of those pages that shows while on the relevant pages.

kazk:
> I think "Getting started" does not need to cover things like translating or authoring a kata, or kumite, because these are not "first contact" activities. I focused on things needed for users being on-boarded.

Completely agree. I want new users to focus on solving and getting used to the platform.

hobovsky:
Glad you like it, if you have any suggestion, just post here.

I will think how to get moving with this, but honestly, I do not even have an idea how to start with this doc system :( I would need some help with questions like:
 - where to put the files (main/toc and subpages),
 - some template to have sidebar, navigation and other stuff available

Knowing this, it will be much easier for me to start and will save us a couple of bogus commits just to "check if it works and what's still missing".

[...]

PR #59, opened 2020-07-16
"Getting started" welcome page

hobovsky:
#40 

Welcome page for "Getting started" tutorial.
Frankly, I did not have a good idea for it. Any suggestions are welcome.


content/getting-started/index.md
```diff
@@ -1,12 +1,14 @@
 ---
 sidebar: getting-started
-next: "/getting-started/setting-up/"
+next: "/getting-started/registering/"
```
kazk, line 3: This and excluding from the sidebar was intentional because I think most of the readers don't need that page and just gets in the way.

> Anyone can view the docs so it might be worth having a page that can be linked from the getting started page (e.g., "If you don't have an account yet, see ...").
> https://github.com/codewars/docs/issues/40#issuecomment-657816008

content/getting-started/index.md
```diff
@@ -1,12 +1,14 @@
...
-TODO Summarize what the reader can learn by reading "Getting Started" pages.
-TODO Add a link to creating a new account at the end of the page just in case the reader haven't registered.
--->
+Welcome to Codewars!
+
+Codewars is a platform which helps you learn, train, and improve your coding skills by solving programming tasks of many types and difficulty levels. You can choose how you would like to learn: do you want to face challenges of increasing difficulty? Or maybe you prefer training by repetition and improving your solutions? After solving a task, you can compare your answer with other users and learn from them too, or you can help less experienced users and answer their questions.
+
+Going through this tutorial, you will learn how to create your new account on Codewars, set it up for optimal training experience, and get familiar with basic elements of user interface. Very soon you will be able to face challenges collected in Codewars library and start learning!
+
+If you haven't joined yet, go to [Sign Up](https://www.codewars.com/join) page and continue reading this tutorial to get you ready.
```
kazk, line 14: This sentence should link to the "Registering" doc and the next navigation should go to the account set up. It's awkward to have this sentence and the next document is about registering.

content/getting-started/index.md
hobovsky: I think it should at least have its entry in the sidebar, because having only inline link in one of articles makes this page very hidden and virtually impossible to find if anyone would, for some reason, need to. But it's okay for me remove both links too, if you want to 👍

kazk: We can leave it in the sidebar because that doesn't get in the way and helps with discoverability. The search should reveal the page as well.

content/getting-started/index.md
```diff
@@ -5,8 +5,10 @@ next: "/getting-started/setting-up/"
...
-TODO Summarize what the reader can learn by reading "Getting Started" pages.
-TODO Add a link to creating a new account at the end of the page just in case the reader haven't registered.
--->
+Welcome to Codewars!
+
+Codewars is a platform which helps you learn, train, and improve your coding skills by solving programming tasks of many types and difficulty levels. You can choose how you would like to learn: do you want to face challenges of increasing difficulty? Or maybe you prefer training by repetition and improving your solutions? After solving a task, you can compare your answer with other users and learn from them too, or you can help less experienced users and answer their questions.
+
+Going through this tutorial, you will learn how to create your new account on Codewars, set it up for optimal training experience, and get familiar with basic elements of user interface. Very soon you will be able to face challenges collected in Codewars library and start learning!
+
+If you haven't joined yet, read how to [register](./registering/) your new account on Codewars.
```
kazk, line 14:
```suggestion
If you haven't joined yet, read how to [register](/getting-started/registering/) your new account on Codewars.
```

This works because `index.md` is special (it becomes the parent `/getting-started/`), but I think it's better to be consistent with the rest of the pages.

PR #62, opened 2020-07-17
Add Markdown reference stub and document Codewars extensions

### LINKED ISSUE #27: Document Math typesetting extension for kata descriptions
I guess everything is in the title...

kazk:
Closes #27


Blind4Basics:
Hi,

it's classified as "reference", tho several things are missing:

* codeblocks: we have currently rendering troubles on cw when the code blocks aren't preceeded and followed by an empty line. This should be mentionned
* about the math part, only the syntax to "declare them" is given, but nothing is said about how to write the maths inside (=> an external link to some documentation, maybe?)

content/references/markdown/extensions.md
```diff
@@ -0,0 +1,130 @@
...
+Only the code block of the active language is shown.
+So if the active language was TypeScript, the above Markdown outputs:
+
+> ```typescript
+> const add = (a: number, b: number) => a + b;
+> ```
+
+This is useful when showing examples.
+
+Note that the code block must have info string set to a valid language id.
```
Blind4Basics, line 38:
```suggestion
Notes:
* The code block must have info string set to a valid language id. To find the language id, visit its documentation from the [languages page](/languages/).
* To avoid rendering errors, always put an empty line before and after the code block or the group of code bloks, while consecutive code blocks for different languages must not be separated with empty lines.
```

content/references/markdown/extensions.md
```diff
@@ -0,0 +1,130 @@
...
+```
+```if-not:ruby
+Hidden _if_ the active language is Ruby.
+```
+````
+
+If the active language is Ruby, the above outputs:
+
+> Hidden _if_ the active language is JavaScript **or** TypeScript.
+
```
Blind4Basics, line 80:
```suggestion
Just like code blocks, language specific blocks needs an empty line before and after them or their group to be rendered correctly.
```

Blind4Basics:
Question: I actually don't really get what are "extensions" for markdowns. Why a separated page while everything was in the same in the original wiki?

kazk:
> codeblocks: we have currently rendering troubles on cw when the code blocks aren't preceeded and followed by an empty line. This should be mentionned

Can you elaborate? We should support Markdown following CommonMark spec. If something isn't working, then it's not a valid Markdown or a Codewars bug.

> about the math part, only the syntax to "declare them" is given, but nothing is said about how to write the maths inside (=> an external link to some documentation, maybe?)

Good point. It's a subset of LaTex and we use [KaTex](https://katex.org/). I'll add a link to it.

> Question: I actually don't really get what are "extensions" for markdowns. Why a separated page while everything was in the same in the original wiki?

Those are added by us and not part of Markdown. It should be in the separate page to make that clear. It also only works for kata descriptions and not useful for the majority of the users.

Blind4Basics:
> elaborate

Currently:

````
txt

```
some stuff in there
```
txt directly
````

this lead to unrendered code block (unless it has changed recently again)

Similarly, there are problems with bullet points recently. [Example](https://www.codewars.com/kata/5252f11b401ddbf7090006c4/).

> extensions

:+1:

**Comment by kazk** [comment]
Thanks. Not showing a code block sounds like a bug, and we should open an issue about it. Do you remember the kata that had this problem? I need to see the exact Markdown that caused it because it can behave differently based on info strings and the context (what's around it).

For bullet points, I'm assuming the problem was the paragraph after the list was included in the list item. That's according to the spec and the original had invalid (CommonMark) Markdown. It used to be rendered differently because the library we use wasn't confirming to CommonMark spec then.

File: content/references/markdown/extensions.md (hunk 1 above)
Comment (kazk) [reply content/references/markdown/extensions.md:38]: Your suggestion looks good, but it includes the text from the next line so committing this will create duplicates.

You can comment on multiple lines by clicking the line number, then Shift+click the ending line. This way, you can suggest changes to multiple lines.

File: content/references/markdown/extensions.md [hunk 3]
```diff
@@ -0,0 +1,130 @@
...
+So if the active language was TypeScript, the above Markdown outputs:
+
+> ```typescript
+> const add = (a: number, b: number) => a + b;
+> ```
+
+This is useful when showing examples.
+
+Note that the code block must have info string set to a valid language id.
+To find the language id, visit its documentation from the [languages page](/languages/).
```
kazk, line 39:
```suggestion
Notes:
- The code block must have info string set to a valid language id. To find the language id, visit its documentation from the [languages page](/languages/).
- To avoid rendering errors, always put an empty line before and after the code block or the group of code bloks, while consecutive code blocks for different languages must not be separated with empty lines.
```

content/references/markdown/extensions.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Conditional Rendering
+
+Similarly, conditional rendering based on the active language is supported with code blocks with info strings `if:` and `if-not:`.
```
kazk, line 43:
```suggestion
Similarly, conditional rendering based on the active language is supported with code blocks with info strings `if:` and `if-not:`.
Just like code blocks, language specific blocks need an empty line before and after them or their group to be rendered correctly.
```

content/references/markdown/extensions.md
kazk: I'll include this at the top of the section instead.

PR #87, opened 2020-07-22
Add Problem Matcher for TextLint

kazk:
This should show the TextLint issues as annotations.

https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md

---

I might need to prevent TextLint from running when autoformat is done.


kazk:
I'm going to try [Vale](https://errata-ai.gitbook.io/vale/) instead.

Issue #102, opened 2020-07-28
Links general policy for the docs?


Blind4Basics:
I'm going through @kask's message on zulip, and I'm wondering if we shouldn't try to only use the `[text][refLink]` syntax: this way, the actual links are stored once and for all at the bottom of the file, so if ever an update is needed, it's way easier to maintain.

Opinions?

Example (the index page for the markdowns):

```markdown
Codewars supports [Markdown][wiki-markdown].

More specifically, [GitHub Flavored Markdown][gfm] (strict superset of [CommonMark][common-mark]) and few [Codewars extensions][extensions].

<!-- the following lines are defining the links and aren't visible in the rendered page -->
[wiki-markdown]: https://en.wikipedia.org/wiki/Markdown
[common-mark]: https://commonmark.org/
[gfm]: https://github.github.com/gfm/
[extensions]: /references/markdown/extensions/
```

I'll update #74 accordingly.

hobovsky:
I wonder if it will be really helpful since every page would contain references anyway. You won't be able to update one definition in some kind of template and have all references across all docs updated, am I right?

Blind4Basics:
mmmh, nope, but at least, you'd have to do the change one single time per link and per page to update (the main reason, imo, is to know that if you have to check something, you just have to go to see at the very hand of the page, without any risk to miss a link somewhere inside the text).

Blind4Basics:
or maybe limit that to external links?

(but, I mean, could be useful for internal links to, if one day we need to change the names of some files... Well, I guess that the best way in the case would actually be to use so `s.replace(//g,m=>...)` on all the files, in that case)

kazk:
I tend to use the referencing when the URLs gets in the way.
```markdown
Codewars supports [Markdown](https://en.wikipedia.org/wiki/Markdown).

More specifically, [GitHub Flavored Markdown](https://github.github.com/gfm/) (strict superset of [CommonMark](https://commonmark.org/)) and few [Codewars extensions](/references/markdown/extensions/).
```
Too many URLs in between words are annoying to me.

For these, I tend to first write without the reference
```markdown
Codewars supports [Markdown][].

More specifically, [GitHub Flavored Markdown][] (strict superset of [CommonMark][]) and few [Codewars extensions][].
```
Those acts as placeholders as I write. Then I'll add the references or just inline the URLs.

Blind4Basics:
so, no special preferences to adopt for the docs, ok. :+1:

PR #105, opened 2020-07-29
Create sql-writting-tests.md

Blind4Basics:
* Is the location of the file ok?
* I completed the frontmatter, but without the sidebar and prev/next links => ok like that?
* I let the extrenal links as they were in the original post (about that, I still wait some more feedback on #102 )

---

line 20: TODO

> ...Include examples, schema images, etc..._

### DIFF
--- a/content/languages/sql/sql-writting-tests.md
+++ b/content/languages/sql/sql-writting-tests.md
@@ -48,14 +48,14 @@ def show_diff_table(actual, expected)
     puts "<TAB:TABLE:Results: Expected>#{expected.to_json()}"
   else
     daff_data = DaffWrapper.new(
-      actual, 
-      expected, 
+      actual,
+      expected,
       index: true
     ).serializable
     Display.daff_table(
-      daff_data, 
-      label: "Diff", 
-      tab: true, 
+      daff_data,
+      label: "Diff",
+      tab: true,
       allow_preview: true
     )
   end
@@ -73,24 +73,24 @@ describe "Query tests" do
       varchar(:name)
     end
   end
-  
+
   it "should work on an example test" do
     DB[:widgets].insert(name: "foo")
     DB[:widgets].insert(name: "quux")
     DB[:widgets].insert(name: "foobar")
-    expected = [{:id => 1, :name => "foo"}, 
+    expected = [{:id => 1, :name => "foo"},
                 {:id => 3, :name => "foobar"}]
     actual = run_sql.to_a()
     show_diff_table(actual, expected)
     expect(actual).to eq(expected)
   end
-  
+
   it "should work on a random test" do
     40.times do
       name = (1..rand(1..4)).map {"foo".chars.sample}.join
       DB[:widgets].insert(name: name)
     end
-    
+
     ref_soln_query = "SELECT * FROM widgets WHERE widgets.name LIKE 'foo%';"
     expected = DB[ref_soln_query].to_a()
     actual = run_sql.to_a()
--- a/content/languages/sql/sql-writting-tests.md
+++ b/content/languages/sql/sql-writting-tests.md
@@ -79,13 +79,13 @@ describe "Query tests" do
     show_diff_table(actual, expected)
     expect(actual).to eq(expected)
   end
-  
+
   it "should work on a random test" do
     40.times do
       name = (1..rand(1..4)).map {"foo".chars.sample}.join
       DB[:widgets].insert(name: name)
     end
-    
+
     ref_soln_query = "SELECT * FROM widgets WHERE widgets.name LIKE 'foo%';"
     expected = DB[ref_soln_query].to_a()
     actual = run_sql.to_a()
--- a/content/languages/sql/sql-writting-tests.md
+++ b/content/languages/sql/sql-writting-tests.md
@@ -7,7 +7,9 @@ tags:
   - testing
 ---
 
-# Generalities
+# Writting tests for SQL
+
+## Generalities
 
 If you do use an existing SQL kata, be sure to avoid exposing the solution in the preloaded code file. It's easy for a code warrior to read the file in the sample tests and view the solution. Also, drop and recreate the database per `it` block so that there's no surprising state shared across cases.
 
--- a/content/languages/sql/sql-writting-tests.md
+++ b/content/languages/sql/sql-writting-tests.md
@@ -1,9 +1,4 @@
----
-kind: tutorial
-languages: [sql]
-tags:
-  - testing
----
+
 
 If you do use an existing SQL kata, be sure to avoid exposing the solution in the preloaded code file. It's easy for a code warrior to read the file in the sample tests and view the solution. Also, drop and recreate the database per `it` block so that there's no surprising state shared across cases.
 
--- a/content/languages/sql/sql-writting-tests.md
+++ b/content/languages/sql/sql-writting-tests.md
@@ -1,3 +1,14 @@
+---
+description: ""
+kind: tutorial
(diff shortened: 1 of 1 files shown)


content/languages/sql/sql-writting-tests.md
```diff
@@ -0,0 +1,81 @@
+---
+description: ""
+kind: tutorial
+languages: [sql]
+topic: tests
+tags:
+  - testing
```
kazk, line 7:
```suggestion
  - testing
sidebar: "language:sql"
prev: /languages/sql/
```

`sidebar: "language:id"` will show all the pages related to the language.

Blind4Basics: damn it, I didn't see this message before. 

Isn't there a way to get notifications for the reviews? (maybe it's sent by email? Then I'd have to change the one used for github... :/ )

Q: should we have a list of "topics" as we do for "kinds" or so? Because we might end up with `test/tests/testing` or so, at some point.

Blind4Basics:
the review above being closed, I duplicate the message here:

Isn't there a way to get notifications for the reviews? (maybe it's sent by email? Then I'd have to change the one used for github... :/ )

Q: should we have a list of "topics" as we do for "kinds" or so? Because we might end up with `test/tests/testing` or so, at some point.

Blind4Basics:
btw, what about that:

> line 20: TODO
>
>    ...Include examples, schema images, etc..._

we let the page as it is for now? (this might rather belong to a tutorial rather than to the reference?)

kazk:
> Isn't there a way to get notifications for the reviews? 

I only get notification on GitHub. We can add to Zulip if you want.

> Q: should we have a list of "topics" for the frontmatter, as we do for "kinds" or so? 

We don't have `Topic` defined yet. Feel free to open an issue for collecting them.

> we let the page as it is for now? (this might belong to a tutorial rather than to the reference?)

We can always improve the docs later so it's up to you.
This teaches someone how to write tests for SQL kata so it's a tutorial. Reference will be a list of available methods.

content/languages/sql/sql-writting-tests.md
```diff
@@ -0,0 +1,83 @@
...
+sidebar: "language:sql"
+prev: /languages/sql/
+---
+
+# Writting tests for SQL
+
+## Generalities
+
+If you do use an existing SQL kata, be sure to avoid exposing the solution in the preloaded code file. It's easy for a code warrior to read the file in the sample tests and view the solution. Also, drop and recreate the database per `it` block so that there's no surprising state shared across cases.
```
kazk, line 16: This is weird as a first sentence on this page.

content/languages/sql/sql-writting-tests.md
```diff
@@ -0,0 +1,83 @@
...
+kind: tutorial
+languages: [sql]
+topic: tests
+tags:
+  - testing
+sidebar: "language:sql"
+prev: /languages/sql/
+---
+
+# Writting tests for SQL
```
kazk, line 12:
```suggestion
# Writing explicit tests for SQL kata
```

content/languages/sql/sql-writting-tests.md
```diff
@@ -0,0 +1,83 @@
...
+
+Disadvantages of this approach are that the challenger might be surprised or confused by the Ruby code, so offering an explanatory note at the end of the instructions can go a long way to mitigating this. Another possible downside is that the test file can become verbose, so you can strategically move some of the content to helper functions in the preloaded file.
+
+Other than Codewars' `run_sql` function, the optional diff table code and interactions with the [DB object](https://github.com/jeremyevans/sequel#label-A+Short+Example), the code uses normal RSpec comparisons on an array of hashes.
+
+Random tests are important to include for the submission to prevent cheating and can use [Faker](https://github.com/faker-ruby/faker) and Ruby's [Random](https://ruby-doc.org/core-2.4.0/Random.html) class.
+
+## Example instructions:
```
kazk, line 26:
```suggestion
## Example instructions
```

content/languages/sql/sql-writting-tests.md
```diff
@@ -0,0 +1,83 @@
...
+
+Write a query to retrieve all rows from the `widgets` table where the varchar widget `name` column starts with the substring `"foo"`. _...Include examples, schema images, etc..._
+
+## Complete solution:
```
kazk, line 30:
```suggestion
## Complete solution
```

content/languages/sql/sql-writting-tests.md
```diff
@@ -0,0 +1,83 @@
...
+
+```sql
+SELECT * FROM widgets WHERE widgets.name LIKE 'foo%';
+```
+
+## Initial solution:
```
kazk, line 36:
```suggestion
## Initial solution
```

content/languages/sql/sql-writting-tests.md
```diff
@@ -0,0 +1,83 @@
...
+
+```sql
+-- Write your query here
+```
+
+## Preloaded code:
```
kazk, line 42:
```suggestion
## Preloaded code
```

content/languages/sql/sql-writting-tests.md
```diff
@@ -0,0 +1,83 @@
...
+  Display.daff_table(
+    daff_data,
+    label: "Diff",
+    tab: true,
+    allow_preview: true
+  )
+end
+```
+
+## Test cases and example test cases:
```
kazk, line 60:
```suggestion
## Test cases
```

content/languages/sql/sql-writting-tests.md
Blind4Basics: I'll check that one tomorrow (too much tired for that now. It's actually the original text, I just updated the post of the issue to make it compatible with the docs.

Blind4Basics:
> I only get notification on GitHub. We can add to Zulip if you want. ([codewars/zulip-github-events#3](https://github.com/codewars/zulip-github-events/issues/3))

Yup, would be cool, imo.


> > Q: should we have a list of "topics" for the frontmatter, as we do for "kinds" or so?
> 
> We don't have `Topic` defined yet. Feel free to open an issue for collecting them.

done (with a question already... XD )

content/languages/sql/sql-writting-tests.md
Blind4Basics: I tried something... Can you check again, plz?

content/languages/sql/sql-writting-tests.md
```diff
@@ -0,0 +1,96 @@
...
+
+Disadvantages of this approach are that the challenger might be surprised or confused by the Ruby code, so offering an explanatory note at the end of the instructions can go a long way to mitigating this. Another possible downside is that the test file can become verbose, so you can strategically move some of the content to helper functions in the preloaded file.
+
+Other than Codewars' `run_sql` function, the optional diff table code and interactions with the [DB object](https://github.com/jeremyevans/sequel#label-A+Short+Example), the code uses normal RSpec comparisons on an array of hashes.
+
+Random tests are important to include for the submission to prevent cheating and can use [Faker](https://github.com/faker-ruby/faker) and Ruby's [Random](https://ruby-doc.org/core-2.4.0/Random.html) class.
```
Blind4Basics, line 22:
```suggestion
Random tests are important to include for the submission to prevent cheating and can use [Faker](https://github.com/faker-ruby/faker) and Ruby's [Random](https://ruby-doc.org/core-2.4.0/Random.html) class to generate mock data. Then run the reference query on the database and assert the comparison of the result arrays as usual.
```

PR #138, opened 2020-10-30
cpp: stringizers

hobovsky:
Examples on using stringizers so assertion messages in C++ tests are more meaningful.


DonaldKellett (approved):

hobovsky:
To save you some clicking, here's the deploy preview of the page: https://deploy-preview-138--reverent-edison-2864ea.netlify.app/languages/cpp/igloo/stringizers/

Would appreciate any review.

Thanks!

content/languages/cpp/igloo/stringizers.md
```diff
@@ -65,25 +65,200 @@ Test output:
 ```
   does_not_pretty_print_vector_of_type_without_stringizer
     Expected: equal to [ [unsupported type], [unsupported type] ]
-    Actual: [ [unsupported type] ]
+    Actual: [ [unsupported type], [unsupported type] ]
 ```
 
 Not only your custom types can be affected by this issue. Many built-in, standard, or 3rd party types also cannot be stringified. Basically, every type which does not define its version of output stream operator (`operator <<`) is affected, and `std::pair` template is a very common case of such type for Codewars kata.
```
error256, line 71: If they "cannot be stringified", we wouldn't be able to stringify them either.

error256 (changes requested):
Looks mostly OK, other than grammatical mistakes.

content/languages/cpp/igloo/stringizers.md
hobovsky:
```suggestion
Not only your custom types can be affected by this issue. Snowhouse may be not able to stringify many built-in, standard, or 3rd party types. Basically, every type which does not define its version of output stream operator (`operator <<`) is affected, and `std::pair` template is a very common case of such type for Codewars kata.
```

content/languages/cpp/igloo/stringizers.md
```diff
@@ -8,3 +8,438 @@ tags:
 ---
 
 # Adding Custom Stringizers
+
+**NOTE:** code snippets in this article omit `#inlude` directives for brevity. You must remember about including required header files!
```
hobovsky, line 12:
```suggestion
**NOTE:** code snippets in this article omit `#include` directives for brevity. You must remember about including required header files!
```

content/languages/cpp/igloo/stringizers.md
```diff
@@ -8,3 +8,438 @@ tags:
...
+
+```
+should_pretty_print_vector_of_pairs_of_custom_type
+  Expected: equal to [ (A, (0, 0)), (B, (2, 2)) ]
+  Actual: [ (A, (1, 1)) ]
+```
+
+## Additional info
+ - [Getting better output for your types](https://github.com/banditcpp/snowhouse#getting-better-output-for-your-types) in Snowhouse repository.
+
```
hobovsky, line 253: Add link to kumite?

https://www.codewars.com/kumite/5f9b749a9cd7a8002dfa2f34/

content/languages/cpp/igloo/stringizers.md
```diff
@@ -8,3 +8,438 @@ tags:
...
+    Actual: (0, 0)
+  should_pretty_print_vector_of_elements_of_custom_type_with_ostream_operator
+    Expected: equal to [ (0, 0), (2, 2) ]
+    Actual: [ (1, 1) ]
+```
+
+
+### Stringification with `Stringizer<T>`
+
+Sometimes, definition of `operator <<` cannot be used for stringification. You cannot add it to types which you do not own, or you cannot re-define it types which already have it defined, but not in a way you'd like to. Or maybe you just prefer this way rather than `operator <<`. In such cases, Snowhouse framework allows you to use a `snowhouse::Stringizer<T>` template specialized for the type you want to stringify.
```
hobovsky, line 136:
```suggestion
Sometimes, definition of `operator <<` cannot be used for stringification. You cannot add it to types which you do not own, or you cannot re-define it for types which already have it defined, but not in a way you'd like to. Or maybe you just prefer this way rather than `operator <<`. In such cases, Snowhouse framework allows you to use a `snowhouse::Stringizer<T>` template specialized for the type you want to stringify.
```

error256 (approved):

kazk (approved):

content/languages/cpp/igloo/stringizers.md
hobovsky:
```suggestion
 - [Kumite](https://www.codewars.com/kumite/5f9b749a9cd7a8002dfa2f34?sel=5f9b749a9cd7a8002dfa2f34) which illustrates how to use stringification in a kata.
```

DonaldKellett (approved):

Issue #8, opened 2020-07-07
Kinds of Documentation


kazk:
We should write documentation with purpose and target audience in mind.

https://github.com/codewars/docs/blob/409a7cd9482b363e8536266d7f15a1dd3ae66331/data/kinds.yml#L9-L36

<img alt="The documentation system" src="https://user-images.githubusercontent.com/639336/87497869-4dc64200-c60b-11ea-9a9d-ee21d20bca07.png" width="600px">

Reference: https://documentation.divio.com/

kazk:
@hobovsky @Blind4Basics 

I decided to name it `Kind` instead of `DocType`. See [data/kinds.yml](https://github.com/codewars/docs/blob/master/data/kinds.yml). The main 4 kinds were taken from https://documentation.divio.com/.

I think this system is great and would like to follow this. What do you guys think?

Blind4Basics:
sounds good. I guess we might sometimes have a hard time to choose between recipes and tutorials but... x)

kazk:
Yeah, it's difficult and the article explains that's why writing a good documentation is difficult. I'd recommend reading https://documentation.divio.com/introduction, but here's my understanding:

Recipes (How-to guides in the linked article) show how to solve a specific goal. The reader isn't new to the topic itself, but looking for a solution for a problem within it. Recipes should present the problem and the solution for it. It can also include some discussions on the problem.

Tutorials are for newcomers. We need to familiarize the reader with the topic. We need to walk them through all the necessary steps and let them learn how things work. Tutorials should only include the necessary information for each step to prevent overwhelming the reader.

- Recipe: "How to write random tests in JavaScript"
- Tutorial: "Creating Your First Kata"

PR #189, opened 2020-12-16
Python: Authoring kata 

hobovsky:
Discussion: #191 
Deployment preview: [here](https://deploy-preview-189--reverent-edison-2864ea.netlify.app/languages/python/authoring/)


content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,213 @@
...
+See the linked issue for details and possible workarounds.
+:::
+
+
+### Input mutation
+
+Issues caused by input mutation are particularly difficult to deal with, because it can lead to bugs which are very subtle, confusing, and difficult to diagnose. When the input is mutated in uncontrolled way, tests may sometimes appear to randomly crash, give incorrect results, or produce confusing logs and assertion messages. Unfortunately, instances of many commonly used data types and classes, are mutable. To avoid problematic situations, following precautions should be taken:
+- Ideally, inputs should be immutable (but unfortunately it's not always possible). Otherwise,
+- Requirements on the mutation of input should be always specified in description _and enforced_. If user solution is required to not modify the received data, there should be a dedicated test case or assertion for that.
```
Steffan153, line 84:
```suggestion
- Requirements on the mutation of input should be always specified in description _and enforced_. If the user solution is required to not modify the received data, there should be a dedicated test case or assertion for that. 
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,213 @@
...
+:::
+
+
+### Input mutation
+
+Issues caused by input mutation are particularly difficult to deal with, because it can lead to bugs which are very subtle, confusing, and difficult to diagnose. When the input is mutated in uncontrolled way, tests may sometimes appear to randomly crash, give incorrect results, or produce confusing logs and assertion messages. Unfortunately, instances of many commonly used data types and classes, are mutable. To avoid problematic situations, following precautions should be taken:
+- Ideally, inputs should be immutable (but unfortunately it's not always possible). Otherwise,
+- Requirements on the mutation of input should be always specified in description _and enforced_. If user solution is required to not modify the received data, there should be a dedicated test case or assertion for that. 
+- If input mutation is allowed for user solutions, reference solution (if used) should not modify it anyway. If it does, it _must_ receive a (deep) copy of the input. Data which was mutated by a reference solution _must not_ be used in any way afterwards (as an input for user solution, or to compose logs or diagnostic messages, etc).
```
Steffan153, line 85:
```suggestion
- If input mutation is allowed for user solutions, the reference solution (if used) should not modify it anyway. If it does, it _must_ receive a (deep) copy of the input. Data which was mutated by a reference solution _must not_ be used in any way afterwards (as an input for user solution, or to compose logs or diagnostic messages, etc).
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,213 @@
...
+- Assertion message _must not_ be composed from a mutable data which could be potentially modified by a user or reference soltion.  
+- Appropriate assertion function should be used for a given test. `assert_equals` is not suitable in all situations. Use `assert_approx_equals` for floating point comparisons, `expect` for tests on boolean values, `expect_error` to test error handling.
+- Some additional attention should be paid to the order of parameters passed to assertion functions. It differs between various assertion libraries, and it happens to be quite often confused by authors, mixing up `actual` and `expected` in assertion messages. For Python Testing framework, the order is `(actual, expected)`.
+- One somewhat distinctive feature of Python assertions is that by default, a failed assertion does not cause a test case to fail early. It can lead to unexpected crashes when an actual value had already been asserted to be invalid, but execution of the current test case was not stopped and following assertions continue to refer to it. This behavior can be ovrriden by passing `allow_raise=True` argument to the assertion functions which support it.
```
Steffan153, line 106:
```suggestion
- One somewhat distinctive feature of Python assertions is that by default, a failed assertion does not cause a test case to fail early. It can lead to unexpected crashes when an actual value had already been asserted to be invalid, but the execution of the current test case was not stopped and following assertions continue to refer to it. This behavior can be overridden by passing the `allow_raise=True` argument to the assertion functions which support it.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+## Additional restrictions
+
+### Accessing solution file
+
+Some kata (for example, code golf challenges or anti-cheat tests) would like to access and read user solution file as text. It's possible and can be done by reading a file located at `/workspace/solution.txt`.
```
Steffan153, line 113:
```suggestion
Some kata (for example, code-golf challenges or anti-cheat tests) would like to access and read the user solution file as text. It's possible and can be done by reading a file located at `/workspace/solution.txt`.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Blocking modules
+
+_TBD_
+
+
+## Example test suite
+
+Below you can find an example test suite which covers most common scenarios mentioned in this article. Note that it does not present all possible techniques, so actual test suites can use different structure, as long as they keep to established conventions and do not violate authoring guidelines.
```
Steffan153, line 122:
```suggestion
Below you can find an example test suite that covers most of the common scenarios mentioned in this article. Note that it does not present all possible techniques, so actual test suites can use a different structure, as long as they keep to established conventions and do not violate authoring guidelines.
```

Steffan153 (changes requested):
---
kind: recipe
languages: [python]
sidebar: "language:python"
---

# Creating a Python Kata

This article is meant as help for kata authors and translators who would like to create new content in Python programming language. It attempts to explain how to create and organize things in a way conforming to [authoring guidelines](/authoring/guidelines/), what are the most common pitfalls, and how to avoid them.

This article is not a standalone tutorial on creating a kata or translation. It's meant to be a complementary, Python-specific part to more general set of HOWTOs and guidelines related to [content authoring](/authoring/). 


## General info

Any information related to Python setup in context of Codewars, language version, available modules, and setup of the code runner, can be found on [Python reference](/languages/python/) page.


## Description

Python code blocks can be inserted with Python-specific part in [sequential code blocks](/references/markdown/extensions/#sequential-code-blocks):

~~~
```python
...your code here...
```
~~~

Python-specific paragraphs can be inserted with [language conditional rendering](/references/markdown/extensions/#conditional-rendering):

```
~~~if:python
...text visible only for Python description...
~~~
~~~if-not:python
...text not visible in Python description...
~~~
```

## Tasks and Requirements

Some language constructs and features available in Python do not translate well into other languages, and should be avoided if possible:
- Dynamic typig (ab)used in not recommended ways, for example mixed return types (_"Return result, or string 'Error' if no result can be found."_).

Some kata should not be translated into Python, because it can be difficult to keep their initial idea:
- Python standard library is very rich and has many utilities available  (for example, `itertools`, combinatorics functions, `numpy`), so some requirements become very easy to implement, not matching the rank of the kata,
- Python supports big integers natively, so kata which rely on implementation of arbitrary precision integer arithmetic become trivial in Python.


## Coding

Python code should stick to generally recognized Python conventions, with [PEP-8](https://www.python.org/dev/peps/pep-0008/) being most widely accepted.


## Tests

### Testing framework

Python kata use [Codewars Python testing framework](/languages/python/codewars-test/) to implement and execute tests. You should read its reference page to find out how to use `describe` and `it` blocks for [organization and grouping](/languages/python/codewars-test/#grouping-tests), what are available [assertions](/languages/python/codewars-test/#assertions), etc. 

You should notice that Python testing framework produces one test output entry per assertion, so test output panel can get very noisy.

### Random utilities

Python has a rich [random library](https://docs.python.org/3.8/library/random.html), which can be used to easily generate random integers in requested ranges, generate floating point numbers, or sample and shuffle collections. Functions available there allow for very convenient construction of various random input generators.

:::warning
Python runner is currently affected by a performance issue (reported as [codewars/runner#58](https://github.com/codewars/runner/issues/58)), which sometimes causes generation of large amounts of random numbers to be noticeably slower. Majority of kata should not be affected by it in a meaningful way, but it can sometimes be a problem for performance tests generating large, random sets of data.  
See the linked issue for details and possible workarounds.
:::


### Input mutation

Issues caused by input mutation are particularly difficult to deal with, because it can lead to bugs which are very subtle, confusing, and difficult to diagnose. When the input is mutated in uncontrolled way, tests may sometimes appear to randomly crash, give incorrect results, or produce confusing logs and assertion messages. Unfortunately, instances of many commonly used data types and classes, are mutable. To avoid problematic situations, following precautions should be taken:
- Ideally, inputs should be immutable (but unfortunately it's not always possible). Otherwise,
- Requirements on the mutation of input should be always specified in description _and enforced_. If user solution is required to not modify the received data, there should be a dedicated test case or assertion for that. 
- If input mutation is allowed for user solutions, reference solution (if used) should not modify it anyway. If it does, it _must_ receive a (deep) copy of the input. Data which was mutated by a reference solution _must not_ be used in any way afterwards (as an input for user solution, or to compose logs or diagnostic messages, etc).
- Input which could be potentially modified by a user solution _must not_ be used afterwards. It must not be used as an input for the reference solution, to compose diagnostic messages, or anything else. If necessary, a (deep) copy should be created and passed to the user solution.

### Reference solution

If the test suite happens to use a reference solution to calculate expected values (what [should be avoided](/authoring/guidelines/submission-tests/#reference-solution), but is not always possible), or some kind of reference data, precalculated arrays, etc., it can be redefined, overwritten, or accessed directly by user solution. To avoid this, it should be defined in a non-global scope, local to the testing function, `it` block, or `describe` block. The reference solution or data _must not_ be defined in [Preloaded code](/authoring/guidelines/preloaded/).

### Calling assertions

Python testing framework provides a set of useful [assertions](/languages/python/codewars-test/#assertions), but when used incorrectly, they can cause a series of problems:
- Stacktraces of a crashing user solution can reveal details which should not be visible,
- Reference soluton and user solution called in wrong order can mutate input in undesirable way allowing for abuse or leading to bugs,
- Use of an assertion not suitable for given case can cause incorrect test results,
- Incorrectly used assertion can produce confusing or unhelpful messages.

To avoid above problems, calls to assertion functions should respect some rules:
- Expected and actual values should be calculated _before_ a call to assertion method. Parameters passed to the assertion should not be function call expressions, but values calculated directly before.
- If a reference solution is used to calculate expected value, it _must_ be either called _before_ user solution, or receive a (deep) copy of input data.
- Assertion message _must not_ be composed from a mutable data which could be potentially modified by a user or reference soltion.  
- Appropriate assertion function should be used for a given test. `assert_equals` is not suitable in all situations. Use `assert_approx_equals` for floating point comparisons, `expect` for tests on boolean values, `expect_error` to test error handling.
- Some additional attention should be paid to the order of parameters passed to assertion functions. It differs between various assertion libraries, and it happens to be quite often confused by authors, mixing up `actual` and `expected` in assertion messages. For Python Testing framework, the order is `(actual, expected)`.
- One somewhat distinctive feature of Python assertions is that by default, a failed assertion does not cause a test case to fail early. It can lead to unexpected crashes when an actual value had already been asserted to be invalid, but execution of the current test case was not stopped and following assertions continue to refer to it. This behavior can be overriden by passing `allow_raise=True` argument to the assertion functions which support it.


## Additional restrictions

### Accessing solution file

Some kata (for example, code golf challenges or anti-cheat tests) would like to access and read user solution file as text. It's possible and can be done by reading a file located at `/workspace/solution.txt`.

### Blocking modules

_TBD_


## Example test suite

Below you can find an example test suite which covers most common scenarios mentioned in this article. Note that it does not present all possible techniques, so actual test suites can use a different structure, as long as they keep to established conventions and do not violate authoring guidelines.

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,213 @@
...
+### Calling assertions
+
+Python testing framework provides a set of useful [assertions](/languages/python/codewars-test/#assertions), but when used incorrectly, they can cause a series of problems:
+- Stacktraces of a crashing user solution can reveal details which should not be visible,
+- Reference soluton and user solution called in wrong order can mutate input in undesirable way allowing for abuse or leading to bugs,
+- Use of an assertion not suitable for given case can cause incorrect test results,
+- Incorrectly used assertion can produce confusing or unhelpful messages.
+
+To avoid above problems, calls to assertion functions should respect some rules:
+- Expected and actual values should be calculated _before_ a call to assertion method. Parameters passed to the assertion should not be function call expressions, but values calculated directly before.
```
DonaldKellett, line 101:
```suggestion
- Expected and actual values should be calculated _before_ invoking an assertion. Parameters passed to the assertion should not be function call expressions, but values calculated directly beforehand.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,213 @@
...
+- If a reference solution is used to calculate expected value, it _must_ be either called _before_ user solution, or receive a (deep) copy of input data.
```
DonaldKellett, line 102:
```suggestion
- If a reference solution is used to calculate the expected value, it _must_ be invoked _before_ the user solution or receive a deep copy of the input data.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,213 @@
...
+- Assertion message _must not_ be composed from a mutable data which could be potentially modified by a user or reference soltion.
```
DonaldKellett, line 103:
```suggestion
- Assertion messages _must not_ be composed from mutable data potentially modified by the user or reference solution.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,213 @@
...
+- Reference soluton and user solution called in wrong order can mutate input in undesirable way allowing for abuse or leading to bugs,
+- Use of an assertion not suitable for given case can cause incorrect test results,
+- Incorrectly used assertion can produce confusing or unhelpful messages.
+
+To avoid above problems, calls to assertion functions should respect some rules:
+- Expected and actual values should be calculated _before_ a call to assertion method. Parameters passed to the assertion should not be function call expressions, but values calculated directly before.
+- If a reference solution is used to calculate expected value, it _must_ be either called _before_ user solution, or receive a (deep) copy of input data.
+- Assertion message _must not_ be composed from a mutable data which could be potentially modified by a user or reference soltion.  
+- Appropriate assertion function should be used for a given test. `assert_equals` is not suitable in all situations. Use `assert_approx_equals` for floating point comparisons, `expect` for tests on boolean values, `expect_error` to test error handling.
```
DonaldKellett, line 104:
```suggestion
- Appropriate assertion functions should be used for each given test. `assert_equals` is not suitable in all situations. Use `assert_approx_equals` for floating point comparisons, `expect` for tests on boolean values, `expect_error` to test error handling.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,213 @@
...
+- Some additional attention should be paid to the order of parameters passed to assertion functions. It differs between various assertion libraries, and it happens to be quite often confused by authors, mixing up `actual` and `expected` in assertion messages. For Python Testing framework, the order is `(actual, expected)`.
```
DonaldKellett, line 105:
```suggestion
- Some additional attention should be paid to the order of parameters passed to assertion functions. It differs between various assertion libraries, and it happens to be quite often confused by authors, mixing up `actual` and `expected` in assertion messages. For the Python testing framework, the order is `(actual, expected)`.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,213 @@
...
+- One somewhat distinctive feature of Python assertions is that by default, a failed assertion does not cause a test case to fail early. It can lead to unexpected crashes when an actual value had already been asserted to be invalid, but the execution of the current test case was not stopped and following assertions continue to refer to it. This behavior can be overridden by passing the `allow_raise=True` argument to the assertion functions which support it.
+
+
+## Additional restrictions
+
+### Accessing solution file
```
DonaldKellett, line 111:
```suggestion
### Accessing the solution file
```

DonaldKellett (changes requested):

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Tasks and Requirements
+
+Some language constructs and features available in Python do not translate well into other languages, and should be avoided if possible:
+- Abuse of dynamic typing, e.g. mixed return types (_"Return the result, or the string 'Error' if no result can be found."_).
+
+Some kata should not be translated into Python, because it can be difficult to keep their initial idea:
+- The Python standard library is very rich and has many utilities available (e.g. `itertools`, combinatorics functions, `numpy`), so some nontrivial requirements in other languages become trivial in Python,
+- Python supports big integers natively, so kata that rely on the implementation of arbitrary precision integer arithmetic become trivial in Python.
+- Codewars runner provide a set of preinstalled packages, which are available for users solving kata. They can be real game changer when it comes to difficulty of a Python transltation. For example, `numpy` can make many matrix manipulation kata much easier.
```
Steffan153, line 55:
```suggestion
- The Codewars runner provides a set of preinstalled packages, which are available for users solving kata. They can be a real game-changer when it comes to the difficulty of a Python translation. For example, `numpy` can make many matrix manipulation kata much easier.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,224 @@
...
+- [`random.random()`](https://docs.python.org/3.8/library/random.html#random.random) - returns the next random floating point number in the range `[0.0, 1.0)`.
+
+:::warning
+The Python runner is currently affected by a performance issue (reported as [codewars/runner#58](https://github.com/codewars/runner/issues/58)) which sometimes causes the generation of large amounts of random numbers to be noticeably slower. The majority of kata should not be affected by it in any significant way, but it can sometimes be a problem for performance tests generating large, random sets of data.
+See the linked issue for details and possible workarounds.
+:::
+
+### Additional packages
+
+Codewars runner provide a set of preinstalled packages, which are available not only for users solving a kata, but can be also used by authors to build tests and genertors of test cases. For example, `numpy` can be used to make generation of matrices easier.
```
Steffan153, line 89:
```suggestion
The Codewars runner provides a set of preinstalled packages, which are available not only for users solving a kata, but can be also used by authors to build tests and generators of test cases. For example, `numpy` can be used to make the generation of matrices easier.
```

Steffan153 (changes requested):

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,224 @@
...
+- Input which could be potentially modified by a user solution _must not_ be used afterwards. It must not be used as an input for the reference solution, to compose diagnostic messages, or anything else. If necessary, a (deep) copy should be created and passed to the user solution.
+
+### Reference solution
+
+If the test suite happens to use a reference solution to calculate expected values (what [should be avoided](/authoring/guidelines/submission-tests/#reference-solution), but is not always possible), or some kind of reference data, precalculated arrays, etc., it can be redefined, overwritten, or accessed directly by the user solution. To avoid this, it should be defined in a scope local to the testing function, `it` block, or `describe` block. The reference solution or data _must not_ be defined in [Preloaded code](/authoring/guidelines/preloaded/).
```
DonaldKellett, line 101:
```suggestion
If the test suite happens to use a reference solution to calculate expected values (which [should be avoided](/authoring/guidelines/submission-tests/#reference-solution), but is not always possible), or some kind of reference data, precalculated arrays, etc., it can be redefined, overwritten, or accessed directly by the user solution. To avoid this, it should be defined in a scope local to the testing function, `it` block, or `describe` block. The reference solution or data _must not_ be defined in [Preloaded code](/authoring/guidelines/preloaded/).
```

DonaldKellett (changes requested):

Steffan153 (approved):

content/languages/python/authoring/index.md
```diff
@@ -115,6 +115,7 @@ To avoid the above problems, calls to assertion functions should respect the fol
...
 - Some additional attention should be paid to the order of parameters passed to assertion functions. It differs between various assertion libraries, and it happens to be quite often confused by authors, mixing up `actual` and `expected` in assertion messages. For the Python testing framework, the order is `(actual, expected)`.
 - One somewhat distinctive feature of Python assertions is that by default, a failed assertion does not cause a test case to fail early. It can lead to unexpected crashes when an actual value had already been asserted to be invalid, but the execution of the current test case was not stopped and following assertions continue to refer to it. This behavior can be overridden by passing the `allow_raise=True` argument to the assertion functions which support it.
+- To avoid unexpected crashes in tests, it's recommended to perform some additional assertions before assuming that the answer returned by user solution has some particular type, form, or value. For example, if the test suite sorts returned list to verify its correctness, explicit assertion should be added to check whether returned object is actually a list, and not, for exaple, `None`.
```
Steffan153, line 118:
```suggestion
- To avoid unexpected crashes in tests, it's recommended to perform some additional assertions before assuming that the answer returned by the user solution has some particular type, form, or value. For example, if the test suite sorts the returned list to verify its correctness, an explicit assertion should be added to check whether the returned object is actually a list, and not, for example, `None`.
```

Steffan153 (changes requested):

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,225 @@
+---
+kind: recipe
```
kazk, line 2:
```suggestion
kind: tutorial
```

Recipe is problem oriented. It should present a problem and provide a solution to it.

This is a tutorial because it's meant to familiarize someone new. It should walk through what's needed to create a Python kata.
I think this document should focus on just the basics and link to further readings. For example, "Accessing the solution file" should be a separate recipe. It's useful for even experienced authors.

See https://github.com/codewars/docs/issues/8#issuecomment-658961005

hobovsky: Just as I mentioned in #191, I will try to extract some common and general parts to other articles, when I create more pages like this one. Next ones I am thinking about are JS (because it's popular) and C (because it causes many problems to authors with memory management) or C++.

Steffan153 (approved):

ggorlen (approved):

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,214 @@
+---
+kind: tutorial
+languages: [python]
+sidebar: "language:python"
+---
+
+# Python: creating and translating a kata
+
+This article is meant as help for kata authors and translators who would like to create new content in Python. It attempts to explain how to create and organize things in a way conforming to [authoring guidelines](/authoring/guidelines/), what are the most common pitfalls, and how to avoid them.
```
DonaldKellett, line 9:
```suggestion
This article is meant as help for kata authors and translators who would like to create new content in Python. It attempts to explain how to create and organize things in a way conforming to [authoring guidelines](/authoring/guidelines/), the most common pitfalls and how to avoid them.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Calling assertions
+
+The Python testing framework provides a set of useful [assertions](/languages/python/codewars-test/#assertions), but when used incorrectly, they can cause a series of problems:
+- Stacktraces of a crashing user solution can reveal details which should not be visible,
+- Use of an assertion not suitable for the given case may lead to incorrect test results,
+- Incorrectly used assertions may produce confusing or unhelpful messages.
+
+To avoid the above problems, calls to assertion functions should respect the following rules:
+- Expected value should be calculated _before_ invoking an assertion. `expected` parameter passed to the assertion should not be a function call expression, but a value calculated directly beforehand.
```
DonaldKellett, line 103:
```suggestion
- The expected value should be calculated _before_ invoking an assertion. The `expected` parameter passed to the assertion should not be a function call expression, but a value calculated directly beforehand.
```

DonaldKellett (changes requested):

Blind4Basics:
I didn't read it fully, but there is a big mistake about `random.shuffle`: this is a procedure that mutates the input array and doesn't return anything. Hence:
- it should be mentionned int the description of the random methods (upper part)
- this line in the test suite example is actually wrong: `arr = random.shuffle([i for i in range(100)])`. It should be:

```python
arr = [i for i in range(100)]
random.shuffle(arr)
```

Blind4Basics:
about the random tests:

* `#non-global reference solution`: actually, with the current setup, the ref solution can be declared anywhere in the test cases part, the user cannot access it in anyway (because, on the contrary of ruby, for example, the files are imported in other files as modules, when executing the tests, hence. So as long as the ref soluiton isn't pushed into the preloaded part, it's ok).

* about the "copying thing" (to avoid input mutation), the current example is ~~showing an anti-pattern~~... ok, I finally noticed the building of the assertion message with the input. => suggesting the following, showing the 2 completely opposite ways (sorry, that's not in a proper "review way"):

```python

    def generate_large_cases():
        #... actual implementation
        pass                                              # <<< remove this one, the comment is enough

    @it('Small inputs')
    def small_inputs():
        
        inputs = generate_small_inputs()
        for input in inputs:

            #call reference solution first.
            #we know it does not mutate the array, so no copy is needed
            expected = reference_sum_array(input)

            #Compute the assertion message, using the input, before 
            #sending it to the user (to avoid mutation of the input troubles):
            msg = f'Input: {input}'

            #call user solution and get actual answer.
            actual = sum_array(input)
            
            #call assertion function. Assertion message uses original,
            #non-modified input.
            test.assert_equals(actual, expected, msg)

    @it('Large random tests')
    def large_random_tests():
        
        large_inputs = generate_large_cases()
        
        for input in large_inputs:
            
            #If you call the user's solution first, you need to do a 
            #copy (potentially deep copy) of the input
            actual = sum_array(input[:])

            #Expected computed with an unmodified input
            expected = reference_sum_array(input)
                     
            test.assert_equals(actual, expected)
```

hobovsky:
@Blind4Basics thanks for taking a look.

- Ad `random.shuffle`: damn, it was a copy/paste error, will fix it...
- Ad scope of reference solution: interesting, I did not know that. But still, is it good practice to make ref solution not enclosed in the testing function (i.e. make it a top level function of the test snippet)? Is making ref solution local a recommended way, or is it totally irrelevant? has this setup chance to change/break in the future, exposing unscoped ref solution? @kazk any insight on how runner setup might affect this behavior? Sorry but I am not an expert on Python internals.
- Ad example: I think on reworking it, making it completely abstract and unrelated to any actual problem. Using specific problems  makes some presented techniques appear out of place.

kazk:
Python used to get concatenated into a single file and that's why the solution had access to the reference solution and anything defined in the tests. We'll never go back to that.

If you're still worried, you can do more like naming the reference solution with leading underscore and limiting its scope, to make it more difficult to access it.

kazk:
I'd recommend importing solution/preloaded explicitly.

Import the solution explicitly, like:

```python
import solution
from solution import example
```

If it's missing, the runner prepends `from solution import *` for backwards compatibility.

If there's a preloaded module, the solution and the test can import it explicitly as well. Currently, `from preloaded import *` is prepended to the solution file for backwards compatibility.

Since I took over the project, I've been trying to minimize/remove any preprocessing done by the runner. This way, it'll work predictively and it's easier to reproduce the environment locally.

DonaldKellett (approved):

content/languages/python/authoring/index.md
```diff
@@ -56,9 +56,24 @@ Some kata should not be translated into Python, because it can be difficult to k
...
+
+### Code style
+
 Python code should stick to generally recognized Python conventions, with [PEP-8](https://www.python.org/dev/peps/pep-0008/) being most widely accepted.
 
 
+### Imports
+
+It is strongly recommended to import solution modules explicitly. Possible modules are:
+- `solution`, containing user's submitted solution,
```
Steffan153, line 68:
```suggestion
- `solution`, containing the user's submitted solution,
```

content/languages/python/authoring/index.md
```diff
@@ -56,9 +56,24 @@ Some kata should not be translated into Python, because it can be difficult to k
...
+- `preloaded`, containing preloaded code,
+- `codewars_test`, containing [Codewars Python testing framework](/languages/python/codewars-test/).
```
Steffan153, line 70:
```suggestion
- `codewars_test`, containing the [Codewars Python testing framework](/languages/python/codewars-test/).
```

content/languages/python/authoring/index.md
```diff
@@ -56,9 +56,24 @@ Some kata should not be translated into Python, because it can be difficult to k
...
+
+When any of these imports is missing, then, for compatibility reasons, it will be automatically added by Codewars preprocessor when solution is built. However, relying on this behavior is discouraged, as it's considered deprecated and might change in the future.
```
Steffan153, line 72:
```suggestion
When any of these imports are missing, then, for compatibility reasons, it will be automatically added by Codewars preprocessor when the solution is built. However, relying on this behavior is discouraged, as it's considered deprecated and might change in the future.
```

content/languages/python/authoring/index.md
```diff
@@ -90,7 +105,7 @@ The Codewars runner provides a set of preinstalled packages, which are available
 
 ### Reference solution
 
-If the test suite happens to use a reference solution to calculate expected values (which [should be avoided](/authoring/guidelines/submission-tests/#reference-solution), but is not always possible), or some kind of reference data, precalculated arrays, etc., it can be redefined, overwritten, or accessed directly by the user solution. To avoid this, it should be defined in a scope local to the testing function, `it` block, or `describe` block. The reference solution or data _must not_ be defined in [Preloaded code](/authoring/guidelines/preloaded/).
+If the test suite happens to use a reference solution to calculate expected values (which [should be avoided](/authoring/guidelines/submission-tests/#reference-solution), but is not always possible), or some kind of reference data, precalculated arrays, etc., it can be redefined, overwritten, or accessed directly by the user solution. To avoid this, it should be defined in a scope local to the testing function, `it` block, or `describe` block. The reference solution or data _must not_ be defined in top-level scope of test suite, nor in [Preloaded code](/authoring/guidelines/preloaded/).
```
Steffan153, line 108:
```suggestion
If the test suite happens to use a reference solution to calculate expected values (which [should be avoided](/authoring/guidelines/submission-tests/#reference-solution), but is not always possible), or some kind of reference data, precalculated arrays, etc., it can be redefined, overwritten, or accessed directly by the user solution. To avoid this, it should be defined in a scope local to the testing function, `it` block, or `describe` block. The reference solution or data _must not_ be defined in the top-level scope of the test suite, nor in the [Preloaded code](/authoring/guidelines/preloaded/).
```

Steffan153 (changes requested):

Steffan153:
From what I remember, don't `describe` and `it` have to be `test.describe` and `test.it`? Without it, you are still relying on the runner doing _something_, whatever it is.

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Testing framework
+
+Python kata use the [Codewars Python testing framework](/languages/python/codewars-test/) to implement and execute tests. You should read its reference page to find out how to use `describe` and `it` blocks for [organization and grouping](/languages/python/codewars-test/#grouping-tests), what [assertions](/languages/python/codewars-test/#assertions) are available, etc.
+
+You should notice that the Python testing framework produces one test output entry per assertion, so the test output panel can get very noisy.
+
+### Random utilities
+
+Python has a rich [random library](https://docs.python.org/3.8/library/random.html), which can be used to easily generate random integers in requested ranges, generate floating point numbers, or sample and shuffle collections. Functions available there allow for very convenient construction of various random input generators.
```
Steffan153, line 87:
```suggestion
Python has a rich [random library](https://docs.python.org/3.8/library/random.html), which can be used to easily generate random integers in requested ranges, generate floating-point numbers, or sample and shuffle collections. Functions available there allow for very convenient construction of various random input generators.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+
+Some useful functions include:
+- [`random.randrange(stop)`](https://docs.python.org/3.8/library/random.html#random.randrange) - returns a randomly selected element from range `[0, stop)`.
+- [`random.randrange(start, stop[, step])`](https://docs.python.org/3.8/library/random.html#random.randrange) - returns a randomly selected element from a range equivalent to `range(start, stop, step)`.
+- [`random.randint(a, b)`](https://docs.python.org/3.8/library/random.html#random.randint) - returns a random integer `N` such that `a <= N <= b`.
+- [`random.shuffle(x[, random])`](https://docs.python.org/3.8/library/random.html#random.shuffle) - shuffles the sequence `x` in place.
+- [`random.sample(population, k)`](https://docs.python.org/3.8/library/random.html#random.sample) - returns a `k` length list of unique elements chosen from the `population` sequence or set.
+- [`random.random()`](https://docs.python.org/3.8/library/random.html#random.random) - returns the next random floating point number in the range `[0.0, 1.0)`.
```
Steffan153, line 95:
```suggestion
- [`random.random()`](https://docs.python.org/3.8/library/random.html#random.random) - returns the next random floating-point number in the range `[0.0, 1.0)`.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Reference solution
+
+If the test suite happens to use a reference solution to calculate expected values (which [should be avoided](/authoring/guidelines/submission-tests/#reference-solution), but is not always possible), or some kind of reference data, precalculated arrays, etc., it can be redefined, overwritten, or accessed directly by the user solution. To avoid this, it should be defined in a scope local to the testing function, `it` block, or `describe` block. The reference solution or data _must not_ be defined in the top-level scope of the test suite, nor in the [Preloaded code](/authoring/guidelines/preloaded/).
+
+### Calling assertions
+
+The Python testing framework provides a set of useful [assertions](/languages/python/codewars-test/#assertions), but when used incorrectly, they can cause a series of problems:
+- Stacktraces of a crashing user solution can reveal details which should not be visible,
```
Steffan153, line 113:
```suggestion
- Stacktraces of a crashing user solution can reveal details that should not be visible,
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+- Appropriate assertion functions should be used for each given test. `assert_equals` is not suitable in all situations. Use `assert_approx_equals` for floating point comparisons, `expect` for tests on boolean values, `expect_error` to test error handling.
+- Some additional attention should be paid to the order of parameters passed to assertion functions. It differs between various assertion libraries, and it happens to be quite often confused by authors, mixing up `actual` and `expected` in assertion messages. For the Python testing framework, the order is `(actual, expected)`.
+- One somewhat distinctive feature of Python assertions is that by default, a failed assertion does not cause a test case to fail early. It can lead to unexpected crashes when an actual value had already been asserted to be invalid, but the execution of the current test case was not stopped and following assertions continue to refer to it. This behavior can be overridden by passing the `allow_raise=True` argument to the assertion functions which support it.
```
Steffan153, line 121:
```suggestion
- One somewhat distinctive feature of Python assertions is that by default, a failed assertion does not cause a test case to fail early. It can lead to unexpected crashes when an actual value had already been asserted to be invalid, but the execution of the current test case was not stopped and following assertions continue to refer to it. This behavior can be overridden by passing the `allow_raise=True` argument to the assertion functions that support it.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+kind: tutorial
+languages: [python]
+sidebar: "language:python"
+---
+
+# Python: creating and translating a kata
+
+This article is meant as help for kata authors and translators who would like to create new content in Python. It attempts to explain how to create and organize things in a way conforming to [authoring guidelines](/authoring/guidelines/), the most common pitfalls and how to avoid them.
+
+This article is not a standalone tutorial on creating kata or translations. It's meant to be a complementary, Python-specific part of a more general set of HOWTOs and guidelines related to [content authoring](/authoring/) explaining how to create a Python version of a kata in a way which conforms to authoring guidelines and best practices. If you are going to create a Python translation of a kata, or a new kata in Python from scratch, please make yourself familiar with the aforementioned documents related to authoring in general.
```
Steffan153, line 11:
```suggestion
This article is not a standalone tutorial on creating kata or translations. It's meant to be a complementary, Python-specific part of a more general set of HOWTOs and guidelines related to [content authoring](/authoring/) explaining how to create a Python version of a kata in a way that conforms to authoring guidelines and best practices. If you are going to create a Python translation of a kata, or a new kata in Python from scratch, please make yourself familiar with the aforementioned documents related to authoring in general. 
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+# Python: creating and translating a kata
+
+This article is meant as help for kata authors and translators who would like to create new content in Python. It attempts to explain how to create and organize things in a way conforming to [authoring guidelines](/authoring/guidelines/), the most common pitfalls and how to avoid them.
+
+This article is not a standalone tutorial on creating kata or translations. It's meant to be a complementary, Python-specific part of a more general set of HOWTOs and guidelines related to [content authoring](/authoring/) explaining how to create a Python version of a kata in a way which conforms to authoring guidelines and best practices. If you are going to create a Python translation of a kata, or a new kata in Python from scratch, please make yourself familiar with the aforementioned documents related to authoring in general. 
+
+
+## General info
+
+Any information related to the Python setup on Codewars, language version, available modules, and setup of the code runner can be found on the [Python reference](/languages/python/) page.
```
Steffan153, line 16:
```suggestion
Any information related to the Python setup on Codewars, the language version, available modules, and setup of the code runner can be found on the [Python reference](/languages/python/) page.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+### Imports
+
+It is strongly recommended to import solution modules explicitly. Possible modules are:
+- `solution`, containing the user's submitted solution,
+- `preloaded`, containing preloaded code,
+- `codewars_test`, containing the [Codewars Python testing framework](/languages/python/codewars-test/).
+
+When any of these imports are missing, then, for compatibility reasons, it will be automatically added by Codewars preprocessor when the solution is built. However, relying on this behavior is discouraged, as it's considered deprecated and might change in the future.
+
+If a user solution is expected to use functions or classes from `preloaded` module, it's recommended to add required imports to "initial solution" code snippet.
```
Steffan153, line 74:
```suggestion
If a user solution is expected to use functions or classes from the `preloaded` module, it's recommended to add required imports to the "initial solution" code snippet.
```

Steffan153 (approved):

hobovsky:
I missed it, thanks for pointing this out.
The way it was, it would not even work I think.

kazk:
@Steffan153 `describe`/`it` are not defined, so `test.describe`/`test.it` should be used. From Python 3.8, you should explicitly do `import codewars_test as test`. The runner still prepends that for backwards compatibility for now.

There shouldn't be any magic left.

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+~~~
+```
+
+## Tasks and Requirements
+
+Some language constructs and features available in Python do not translate well into other languages, and should be avoided if possible:
+- Abuse of dynamic typing, e.g. mixed return types (_"Return the result, or the string 'Error' if no result can be found."_).
+
+Some kata should not be translated into Python, because it can be difficult to keep their initial idea:
+- The Python standard library is very rich and has many utilities available (e.g. `itertools`, combinatorics functions), so some nontrivial requirements in other languages become trivial in Python,
```
Blind4Basics, line 53:
```suggestion
- The Python standard library is very rich and has many utilities available (e.g. `itertools`, combinatorics functions), so some nontrivial requirements in other languages could become trivial in Python,
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+- Python supports big integers natively, so kata that rely on the implementation of arbitrary precision integer arithmetic become trivial in Python.
```
Blind4Basics, line 54:
```suggestion
- Python supports big integers natively, so kata that rely on the implementation of arbitrary precision integer arithmetic would become trivial in Python.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+- The Codewars runner provides a set of preinstalled packages which are available for users solving the kata. They can be a real game-changer when it comes to the difficulty of a Python translation. For example, `numpy` can make many matrix manipulation kata much easier.
```
Blind4Basics, line 55:
```suggestion
```

'feels redundant with the first point. Amended the first and removed this one.

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Coding
+
+
+### Code style
+
+Python code should stick to generally recognized Python conventions, with [PEP-8](https://www.python.org/dev/peps/pep-0008/) being most widely accepted.
```
Blind4Basics, line 62: I'm not sure to understand that sentence properly (I have troubles because of the "with", I believe)

```suggestion
Python code should stick to generally recognized Python conventions, with [PEP-8](https://www.python.org/dev/peps/pep-0008/) being most widely accepted.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+
+
+### Imports
+
+It is strongly recommended to import solution modules explicitly. Possible modules are:
```
Blind4Basics, line 67:
```suggestion
For further compatibility reasons, it is strongly recommended to import solution modules explicitly.  
Available modules in the context of Codewars are:
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+- `solution`, containing the user's submitted solution,
+- `preloaded`, containing preloaded code,
+- `codewars_test`, containing the [Codewars Python testing framework](/languages/python/codewars-test/).
```
Blind4Basics, line 70: @kazk :
about this, maybe the import of the framework for the test cases part should be done with `importlib.reload` (maybe it's already the case?) Because user might begin to mess with the test functions.

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+If a user solution is expected to use functions or classes from the `preloaded` module, it's recommended to add required imports to the "initial solution" code snippet.
+
+
+## Tests
+
+### Testing framework
+
+Python kata use the [Codewars Python testing framework](/languages/python/codewars-test/) to implement and execute tests. You should read its reference page to find out how to use `describe` and `it` blocks for [organization and grouping](/languages/python/codewars-test/#grouping-tests), what [assertions](/languages/python/codewars-test/#assertions) are available, etc.
+
+You should notice that the Python testing framework produces one test output entry per assertion, so the test output panel can get very noisy.
```
Blind4Basics, line 83:
```suggestion
You should notice that the Python testing framework produces one test output entry per assertion (and even 2 in [some sepcial situations](/languages/python/codewars-test/#timeout-utility)), so the test output panel can get very noisy.
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Random utilities
+
+Python has a rich [random library](https://docs.python.org/3.8/library/random.html), which can be used to easily generate random integers in requested ranges, generate floating-point numbers, or sample and shuffle collections. Functions available there allow for very convenient construction of various random input generators.
+
+Some useful functions include:
```
Blind4Basics, line 89:
```suggestion
Some useful functions include:
* [`random.choices(population[, ...], k=1)`](https://docs.python.org/3.8/library/random.html#random.choices) - extract `k` elements of the population (possibly outputing the same element several times).
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+- [`random.randrange(stop)`](https://docs.python.org/3.8/library/random.html#random.randrange) - returns a randomly selected element from range `[0, stop)`.
+- [`random.randrange(start, stop[, step])`](https://docs.python.org/3.8/library/random.html#random.randrange) - returns a randomly selected element from a range equivalent to `range(start, stop, step)`.
+- [`random.randint(a, b)`](https://docs.python.org/3.8/library/random.html#random.randint) - returns a random integer `N` such that `a <= N <= b`.
```
Blind4Basics, line 92: I'd remove randint, because it's specific to python/ruby, but that's only personal opinion.

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+See the linked issue for details and possible workarounds.
+:::
+
+### Additional packages
+
+The Codewars runner provides a set of preinstalled packages, which are available not only for users solving a kata, but can be also used by authors to build tests and generators of test cases. For example, `numpy` can be used to make the generation of matrices easier.
+
+### Reference solution
+
+If the test suite happens to use a reference solution to calculate expected values (which [should be avoided](/authoring/guidelines/submission-tests/#reference-solution), but is not always possible), or some kind of reference data, precalculated arrays, etc., it can be redefined, overwritten, or accessed directly by the user solution. To avoid this, it should be defined in a scope local to the testing function, `it` block, or `describe` block. The reference solution or data _must not_ be defined in the top-level scope of the test suite, nor in the [Preloaded code](/authoring/guidelines/preloaded/).
```
Blind4Basics, line 108:
```suggestion
If the test suite happens to use a reference solution to calculate expected values (which [should be avoided](/authoring/guidelines/submission-tests/#reference-solution) when possible), or some kind of reference data like precalculated arrays, etc., it must not be possible for the user to redefine, or overwrite it, or to have direct access to its content. To avoid this, it should be defined in a scope local to the testing function, a `it` or a `describe` block.

The reference solution or data ___must not___ be defined in the top-level scope of the test suite, nor in the [Preloaded code](/authoring/guidelines/preloaded/).
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+- One somewhat distinctive feature of Python assertions is that by default, a failed assertion does not cause a test case to fail early. It can lead to unexpected crashes when an actual value had already been asserted to be invalid, but the execution of the current test case was not stopped and following assertions continue to refer to it. This behavior can be overridden by passing the `allow_raise=True` argument to the assertion functions that support it.
+- To avoid unexpected crashes in tests, it's recommended to perform some additional assertions before assuming that the answer returned by the user solution has some particular type, form, or value. For example, if the test suite sorts the returned list to verify its correctness, an explicit assertion should be added to check whether the returned object is actually a list, and not, for example, `None`.
+
+
+## Additional restrictions
```
Blind4Basics, line 125:
```suggestion
## Kata with additional restrictions
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Accessing the solution file
```
Blind4Basics, line 127: there is always that problem about when to do the length check... In some languages you have to do the test in the test cases part to not allow the `JS/return;` cheat, but in python, that's not a problem, afaik. => should we add that the length check must be done in the preloaded section? (so import of the test framewrork there too)

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+
+Some kata (for example, code-golf challenges or anti-cheat tests) would like to access and read the user solution file as text. It's possible and can be done by reading a file located at `/workspace/solution.txt`.
+
+### Blocking modules
+
+_TBD_
+
+
+## Example test suite
```
Blind4Basics, line 136: might be cool to add a seciton in the example where the `it` blocks are built on the fly, inside a loop. That's a commom stuff, when you begin to build more complex test suites.

?

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+def random_tests():
+
+    #non-global reference solution
+    def reference_solution(arr):
+        # calculate and return reference answer
+
+    #generate data for test cases with small inputs
+    #this test case generator combines a few types of input
+    #in one collection
+    def ():
```
Blind4Basics, line 182:
```suggestion
    def generate_small_inputs():    
```

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+
+```python
+
+#import modules explicitly
+import codewars_test as test
+import preloaded
+from solution import user_solution
+
+@test.describe('Fixed tests')
+def fixed_tests():
```
Blind4Basics, line 149:
```suggestion
def _():
```

function names for the blocks are actually useless and redundant with the block message. I generally use `_` instead. Up to you to see if you wanna change them all or none.

content/languages/python/authoring/index.md
hobovsky: I decided to distinguish standard library (or built-in modules, or however it's called in Python) from additional modules, because an author can be aware of, for example, a variety of functions available in built-in `math` but they might be unaware of `scipy` or `numpy` which are installed additionally.

content/languages/python/authoring/index.md
hobovsky, line 67:
```suggestion
For future compatibility reasons, it is strongly recommended to import solution modules explicitly. Possible modules are:
```

content/languages/python/authoring/index.md
hobovsky: I specifically skipped `random.choices`, because I do not see how it can be useful in context of random tests. The fact that it can pick one element multiple times makes it IMO inferior to `random.sample`. Am I missing something here?

content/languages/python/authoring/index.md
hobovsky: But this page _is_ Python specific? I've included it so authors would rather use `randint(a,b)` than `randrange(a, b+1)` or  even `randrange(b-a) + a + 1`. Many languages do not have an equivalent of `randint` available, so authors coming from other background often implement equivalent functions by themselves instead of using existing one.

content/languages/python/authoring/index.md
hobovsky: This part will be moved to more general docs: one on accessing user solution file, and another on how additional restrictions are bad in general. Here I will add something on module forbidder, or totally skip it, IDK yet.

content/languages/python/authoring/index.md
hobovsky: I was thinking about it, did not get any good idea for this. But yes, it wold be worth adding 👍

content/languages/python/authoring/index.md
hobovsky: I'd personally stick to meaningful names, even if they are redundant. I'd be afraid that in larger test suite it will be confusing when you'd end up with functions called `_`, `__`, `___` etc.

Steffan153: It's not a problem to name them all `_` lol, you don't have to go `__` and `___` actually.

hobovsky: So you telling me I can have a test suite which goes like:

```python
@test.describe('Fixed tests')
def _():
    some it blocks here...

@test.describe('Random tests')
def _():
    some it blocks here...
```

Steffan153: Yes. The function name gets overridden ofc, but python doesn't care, and the decorator picks it up in time.

hobovsky: WTF what kind of witchcraft is even that? :D It's definitely not something I would expect to happen.

Steffan153: Lol, Python is a mutable language! XD

hobovsky: Okay, it might be just me, but if I were to guess, I would not expect this, ummm... feature, to not cause problems some time in the future. If anything, if avoiding redundancy would be a goal, I'd make decorators parameterless, and derive the title of the section from decorated function name:

```python
@test.describe
def fixed_tests():
    some it blocks here...

@test.describe
def random_tests():
    some it blocks here...
```

I think I'd document such possibility in the framework reference, in its part on `it` and `describe`, and definitely would not give it as an example ;)

Putting two identical names in the same scope, _willingly_, is asking for trouble - change my mind.

@kazk any insight from technical side, if/how it could affect the runner?

kazk: Python's decorator is a function that takes a function. `@test.describe`/`@test.it` calls the given function. See https://github.com/codewars/python-test-framework/blob/a07c83f0b5a0692f665ee316865510867ab9a4cf/codewars_test/test_framework.py#L103-L121

So it's equivalent to:
```javascript
describe('Fixed Tests', function _() {
});
describe('Random Tests', function _() {
});
```

> Putting two identical names in the same scope, 

I don't think it does that, but I was never interested in Python enough to know the details, so I can be wrong.

kazk: I guess it does define functions `_`. But I don't think it'll do anything bad because overwriting it doesn't change the function passed to the decorator.

kazk:
```python
# Simplified decorator
def factory(output_cmd): # DESCRIBE/IT
    def decorator(name):
        def wrapper(func):
            print(output_cmd, name)
            func()
            print('completedin')
        return wrapper
    return decorator

describe = _factory('DESCRIBE')
@describe('Fixed Tests')
def _():
    # ...
# is a syntactic suger of
def _():
    # ...
_ = _factory('DESCRIBE')('Fixed Tests')(_)
```

So `_` has the return value of `wrapper(func)` which is usually `None`.

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,252 @@
...
+
+
+### Imports
+
+It is strongly recommended to import solution modules explicitly. Possible modules are:
+- `solution`, containing the user's submitted solution,
+- `preloaded`, containing preloaded code,
+- `codewars_test`, containing the [Codewars Python testing framework](/languages/python/codewars-test/).
+
+When any of these imports are missing, then, for compatibility reasons, it will be automatically added by Codewars preprocessor when the solution is built. However, relying on this behavior is discouraged, as it's considered deprecated and might change in the future.
```
DonaldKellett, line 73:
```suggestion
When any of these imports are missing, then, for compatibility reasons, it will be automatically added by Codewars' preprocessor when the solution is built. However, relying on this behavior is discouraged as it's deprecated and subject to change in the future.
```

DonaldKellett (changes requested):

content/languages/python/authoring/index.md
Blind4Basics: about choices, it's about... choices. XD
Sorry... x)

that actually depends what you're building. If you're building data that can contain duplicated values, it's useful.

content/languages/python/authoring/index.md
Blind4Basics: >  I'd make decorators parameterless, and derive the title of the section from decorated function name:

I'd say it would be a bad idea, because doing so, you limit the feedback to valid functions identifiers. Then you'll need to go back to parameters to allow a well formed sentence as feedback, using paremeters again (they did that for JUnit, I beleive. It's just a mess... XD)

content/languages/python/authoring/index.md
Blind4Basics: the module forbidder is python specific, so if it must be talked about, that's here, I guess. I'm still reluctant at looking into that thing, since I've no idea how to resolve the last way Unnamed broke it...

content/languages/python/authoring/index.md
Blind4Basics: I wrote some stuff lik that, I can grab examples...

Blind4Basics: out of the blue examples:

```code
    # loooooong data tests definitions, for TESTS_THE_TOKENIZER1, TESTS_THE_TOKENIZER2, ...
    # and defining the run_tests function that is using a loop to trun all the tests fro each section

    TEST_CONFIG = ( ('should correctly tokenize valid RSU program', TESTS_THE_TOKENIZER1),
                    ('should throw an error if one or more invalid tokens are detected', TESTS_THE_TOKENIZER2),
                    ('should correctly tokenize invalid RSU programs containing only valid tokens', TESTS_THE_TOKENIZER3),
                    ('Additional tests', NEW_TESTS_TOKENIZER),
                   )
    
    for msg,data in TEST_CONFIG:
        @test.it(msg)
        def _():
            run_tests(data)
```

here is the run_tests function, in case you find the intent clearer (not sure it will, because I have two kind of tests in there, changing what is tested depending on `exp` being `None` or not...):

```code
    def run_tests(config_tests):
        for program,exp,msg in config_tests:
            if exp:
                test.assert_equals(..., exp)
            else:
                test.expect_error(msg, lambda: ... )
```

Blind4Basics: I have that one too, but using the timeout thing. Not sure it's a good idea to put it here.

```python

            CONFIG = [
                ((9,9), (17,17), circle(9,9,8)),
                ((0,0), (7,7), ((5,6),(5,8),(6,5),(6,9),(9,6),(8,5),(8,9),(9,8))),
                ((0,0), (9,9), circle(9,9,8)),
            ]
            for s,e,o in CONFIG:
                display(s,e,o)
                @test.timeout(.2)
                def _():
                    ret = _user_func(s,e,o)
                    Test.expect(ret is None,"Unreachable, the result should be None but was: "+str(ret))
```

Blind4Basics: or with more levels:

```python

    msgIts = ["long stuff here", ...]
    for i,(S,E,exp,o) in enumerate([ ((0,14),  (0,35), 31, O),
                                     ((0,-20), (0,20), 24, O),
                                     ((0,-20), (0,35), 35, C1),
                                     ((0,-20), (0,35), 29, C2),
                                   ], 1):
        @Test.describe(f'Case #{i}')
        def _():
        
            timings=[]
            for j,(s,e) in enumerate([(S,E),(E,S)]):
                
                @Test.it(msgIts[j])
                def _():
                    display(s,e,o,limits)
                    act = _user_func(s,e,o)
                    Test.assert_equals(act, exp)
                
            @Test.it("All of that without losing in performances")
            def _():
                limit,a,b = 0.07, *timings
                # ...
                Test.expect(p<limit, f"Variation of execution time should be less than {limit:.1%}")
```

Blind4Basics: I guess the first example is the most appropriate one...

DonaldKellett (approved):

content/languages/python/authoring/index.md
```diff
@@ -0,0 +1,255 @@
...
+- Avoid returning different data types depending on the situation (_"Return the result, or the string 'Error' if no result can be found..."_). Python is dynamically typed which is not be the case for some other languages. Returning `None` might be appropriate in some situations, but raising an exception might be better in others.
+
+Some kata just should not be translated into Python because it can be difficult to keep their initial idea:
+- The Python standard library is very rich and has many utilities available (e.g. `itertools`, combinatorics functions), so some nontrivial requirements in other languages could become trivial in Python,
+- Python supports big integers natively, so kata that rely on the implementation of arbitrary precision integer arithmetic would become trivial in Python.
+- The Codewars runner provides a set of preinstalled packages which are available for users solving the kata. They can be a real game-changer when it comes to the difficulty of a Python translation. For example, `numpy` can make many matrix manipulation kata much easier.
```
Steffan153, line 55: What about a link to https://docs.codewars.com/languages/python/ since it has the list?

hobovsky: It's mentioned at the beginning, you think it's worth repeating here?

Steffan153: ah, didn't see. It would be nice if that was included in the sidebar here tho:
image

hobovsky: Sidebar here is built automatically, i am not sure if it can be modified. I will think about something.

Steffan153: It doesn't show all the links though, apparently only references.

Steffan153: Actually, it can be changed in the `sidebar` property in the gridsome.config.js.

content/languages/python/authoring/index.md
hobovsky: I added a note on `choices`.

content/languages/python/authoring/index.md
hobovsky: Added small example of dynamically generated test cases.

content/languages/python/authoring/index.md
hobovsky: I removed this section. Module forbidded will be described in separate doc, and accessing solution file will be not python-specific.

content/languages/python/authoring/index.md
hobovsky: I added a paragraph on this.

content/languages/python/authoring/index.md
hobovsky: See #195

content/languages/python/authoring/index.md
```diff
@@ -81,7 +81,56 @@ If a user solution is expected to use functions or classes from the `preloaded`
...
+    test_cases = generate_test_cases()
+    for msg, input, expected in test_case:
+        @test.it(msg)
+        def _():
+            actual = user_solution(input)
+            test.assert_equals(actual, expected)
+
+```
+
+This technique is liked by authors familiar with testing frameworks which provide parametrized or generated test cases out of the box, like NUnit, or JUnit. However, some caution is needed when this approach is used. Test suites organized this way can become large and can flood test output panel with many entries, making it unreadable or causing performance problems in client browsers.
```
Steffan153, line 101:
```suggestion
This technique is liked by authors familiar with testing frameworks that provide parametrized or generated test cases out of the box, like NUnit, or JUnit. However, some caution is needed when this approach is used. Test suites organized like this can become large and can flood the test output panel with many entries, making it unreadable or causing performance problems in client browsers.
```

content/languages/python/authoring/index.md
```diff
@@ -81,7 +81,56 @@ If a user solution is expected to use functions or classes from the `preloaded`
...
+
+#### Decorated functions
+
+To create and present test output, Python testing framework uses parameters of `describe` and `it` decorators, and ignores actual names of decorated functions. Since the names are often redundant with titles of `describe` or `it` sections, they can be replaced with some placeholder name, for example `_`:
```
Steffan153, line 105:
```suggestion
To create and present test output, the Python testing framework uses parameters of `describe` and `it` decorators, and ignores actual names of decorated functions. Since the names are often redundant with titles of `describe` or `it` sections, they can be replaced with some placeholder name, for example, `_`:
```

content/languages/python/authoring/index.md
```diff
@@ -91,9 +140,10 @@ Some useful functions include:
...
 - [`random.randint(a, b)`](https://docs.python.org/3.8/library/random.html#random.randint) - returns a random integer `N` such that `a <= N <= b`.
+- [`random.random()`](https://docs.python.org/3.8/library/random.html#random.random) - returns the next random floating-point number in the range `[0.0, 1.0)`.
 - [`random.shuffle(x[, random])`](https://docs.python.org/3.8/library/random.html#random.shuffle) - shuffles the sequence `x` in place.
 - [`random.sample(population, k)`](https://docs.python.org/3.8/library/random.html#random.sample) - returns a `k` length list of unique elements chosen from the `population` sequence or set.
-- [`random.random()`](https://docs.python.org/3.8/library/random.html#random.random) - returns the next random floating-point number in the range `[0.0, 1.0)`.
+- [`random.choices(population[, ...], k=1)`](https://docs.python.org/3.8/library/random.html#random.choices) - extracts `k` elements of the `population` (possibly outputing the same element several times).
```
Steffan153, line 146:
```suggestion
- [`random.choices(population[, ...], k=1)`](https://docs.python.org/3.8/library/random.html#random.choices) - extracts `k` elements of the `population` (possibly outputting the same element several times).
```

Steffan153 (approved):

hobovsky:
Feedback is mostly ok so I am going to merge this PR, and prepare another one on organization, wording and grammar of the framework reference.

Issue #197, opened 2020-12-29
Improve docs on Python testing framework


hobovsky:
After [Python: creating and translating a kata](https://docs.codewars.com/languages/python/authoring/) has been created, docs on [Python testing framework](https://docs.codewars.com/languages/python/codewars-test/) can use some improvements on wording, grammar, and to remove some redundancies.

hobovsky:
I wanted to remove parts which became redundant after introducing "Authoring Python kata" doc, and to make it look more like an actual reference rather than a HOWTO. I hope I did not remove too much, but if you think something is missing, let me know.

PR #198, opened 2020-12-29
Improve docs on Python testing framework

hobovsky:
Discussion: #197 
Deploy preview: [here](https://deploy-preview-198--reverent-edison-2864ea.netlify.app/languages/python/codewars-test/)

I wanted to remove parts which became redundant after introducing "Authoring Python kata" doc, and to make it look more like an actual reference rather than a HOWTO. I hope I did not remove too much, but if you think something is missing, let me know.

### DIFF
--- a/content/languages/python/authoring/index.md
+++ b/content/languages/python/authoring/index.md
@@ -75,6 +75,15 @@ When any of these imports are missing, then, for compatibility reasons, it will 
 
 If a user solution is expected to use functions or classes from the `preloaded` module, it's recommended to add required imports to the "initial solution" code snippet.
 
+:::note
+For simplicity, thorough the rest of this article it's assumed that `codewars_test` module is imported and aliased as `test` with following statement:
+
+```python
+import codewars_test as test
+```
+
+This is a convention used in many Python kata, but it's not a requirement, and authors can choose to import the module in any way they find suitable for them.
+:::
 
 ## Tests
 
@@ -82,10 +91,9 @@ If a user solution is expected to use functions or classes from the `preloaded` 
 
 Python kata use the [Codewars Python testing framework](/languages/python/codewars-test/) to implement and execute tests. You should read its reference page to find out how to use `describe` and `it` blocks for [organization and grouping](/languages/python/codewars-test/#organization-of-tests), what [assertions](/languages/python/codewars-test/#assertions-1) are available, etc.
 
-
 #### Dynamically generated test cases
 
-It's possible to put functions decorated with `it` in a loop and use them as a construct similar to parametrized test cases known from other testing frameworks, for example:
+It's possible to put functions decorated with `@test.it` in a loop and use them as a construct similar to parametrized test cases known from other testing frameworks, for example:
 
 ```python
 @test.describe("Generated test cases")
@@ -103,7 +111,7 @@ This technique is liked by authors familiar with testing frameworks that provide
 
 #### Decorated functions
 
-To create and present test output, the Python testing framework uses parameters of `describe` and `it` decorators, and ignores actual names of decorated functions. Since the names are often redundant with titles of `describe` or `it` sections, they can be replaced with some placeholder name, for example, `_`:
+To create and present test output, the Python testing framework uses parameters of `@test.describe` and `@test.it` decorators, and ignores actual names of decorated functions. Since the names are often redundant with titles of `describe` or `it` sections, they can be replaced with some placeholder name, for example, `_`:
 
 ```python
 @test.describe('Fixed tests')
@@ -170,7 +178,7 @@ The Python testing framework provides a set of useful [assertions](/languages/py
 
 To avoid the above problems, calls to assertion functions should respect the following rules:
 - The expected value should be calculated _before_ invoking an assertion. The `expected` parameter passed to the assertion should not be a function call expression, but a value calculated directly beforehand.
-- Appropriate assertion functions should be used for each given test. `assert_equals` is not suitable in all situations. Use `assert_approx_equals` for floating point comparisons, `expect` for tests on boolean values, `expect_error` to test error handling.
+- Appropriate assertion functions should be used for each given test. `test.assert_equals` is not suitable in all situations. Use `test.assert_approx_equals` for floating point comparisons, `test.expect` for tests on boolean values, `test.expect_error` to test error handling.
 - Some additional attention should be paid to the order of parameters passed to assertion functions. It differs between various assertion libraries, and it happens to be quite often confused by authors, mixing up `actual` and `expected` in assertion messages. For the Python testing framework, the order is `(actual, expected)`.
 - One somewhat distinctive feature of Python assertions is that by default, a failed assertion does not cause a test case to fail early. It can lead to unexpected crashes when an actual value had already been asserted to be invalid, but the execution of the current test case was not stopped and following assertions continue to refer to it. This behavior can be overridden by passing the `allow_raise=True` argument to the assertion functions that support it.
 - To avoid unexpected crashes in tests, it's recommended to perform some additional assertions before assuming that the answer returned by the user solution has some particular type, form, or value. For example, if the test suite sorts the returned list to verify its correctness, an explicit assertion should be added to check whether the returned object is actually a list, and not, for example, `None`.
--- a/content/languages/python/authoring/index.md
+++ b/content/languages/python/authoring/index.md
@@ -80,7 +80,7 @@ If a user solution is expected to use functions or classes from the `preloaded` 
 
 ### Testing framework
 
-Python kata use the [Codewars Python testing framework](/languages/python/codewars-test/) to implement and execute tests. You should read its reference page to find out how to use `describe` and `it` blocks for [organization and grouping](/languages/python/codewars-test/#grouping-tests), what [assertions](/languages/python/codewars-test/#assertions) are available, etc.
+Python kata use the [Codewars Python testing framework](/languages/python/codewars-test/) to implement and execute tests. You should read its reference page to find out how to use `describe` and `it` blocks for [organization and grouping](/languages/python/codewars-test/#organization-of-tests), what [assertions](/languages/python/codewars-test/#assertions-1) are available, etc.
 
 
 #### Dynamically generated test cases
@@ -163,7 +163,7 @@ The reference solution or data ___must not___ be defined in the top-level scope 
 
 ### Calling assertions
 
-The Python testing framework provides a set of useful [assertions](/languages/python/codewars-test/#assertions), but when used incorrectly, they can cause a series of problems:
+The Python testing framework provides a set of useful [assertions](/languages/python/codewars-test/#assertions-1), but when used incorrectly, they can cause a series of problems:
 - Stacktraces of a crashing user solution can reveal details that should not be visible,
 - Use of an assertion not suitable for the given case may lead to incorrect test results,
 - Incorrectly used assertions may produce confusing or unhelpful messages.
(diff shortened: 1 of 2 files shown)


content/languages/python/codewars-test/index.md
```diff
@@ -5,15 +5,59 @@ tags:
...
+
+Additionally, the framework provides a set of functions performing assertions on various conditions: equality, error handling, truthness, etc.
+
+
+## Module `codewars_test`
+
+All functions, decorators, and assertions provided by the framework are defined in `codewars_test` Python module.
+
+:::warning Deprecation
+Python runner for verions prior to Python 3.8 does not contain the `codewars_test` module. The testing framework is imported implicitly and aliased as `test` and `Test`. This behavior is deprecated and for the Python 3.8 kata the explicit import is required.
```
Steffan153, line 27:
```suggestion
The Python runner for versions prior to Python 3.8 does not contain the `codewars_test` module. The testing framework is imported implicitly and aliased as `test` and `Test`. This behavior is deprecated and for the Python 3.8 kata, the explicit import is required.
```

content/languages/python/codewars-test/index.md
```diff
@@ -5,15 +5,59 @@ tags:
...
+:::
+
+## Organization of tests
+
+Tests in Python testing framework are composed from functions decorated with a set of Python decorators. All such functions are automatically discovered and run. The final result for every test block is determined by contained assertions, and reported along with its measured execution time.
```
Steffan153, line 32:
```suggestion
Tests in the Python testing framework are composed of functions decorated with a set of Python decorators. All such functions are automatically discovered and run. The final result for every test block is determined by contained assertions and reported along with its measured execution time.
```

content/languages/python/codewars-test/index.md
```diff
@@ -5,15 +5,59 @@ tags:
...
+
+
+### Test groups: `@describe`
```
Steffan153, line 35: Should be `@test.describe`/`@test.it`. Also below.

content/languages/python/codewars-test/index.md
```diff
@@ -5,15 +5,59 @@ tags:
...
+
+Functions decorated with `@describe` represent a group of logically related test cases.
+
+`@describe` functions can contain another, nested test groups (functions decorated with `@describe`), or individual test cases (functions decorated with `@it`, see below).
```
Steffan153, line 39:
```suggestion
`@describe` functions can contain more nested test groups (functions decorated with `@describe`) or individual test cases (functions decorated with `@it`, see below).
```

content/languages/python/codewars-test/index.md
```diff
@@ -5,15 +5,59 @@ tags:
...
+
+Test groups cannot contain assertions.
+
+
+### Test cases: `@it`
+
+Functions decorated with `@it` represent a single test case. They can be defined inside of a function decorated with `@describe`, or can be top level functions.
+
+`@it` functions can contain only assertions, and cannot nest another test cases nor test groups.
```
Steffan153, line 48:
```suggestion
`@it` functions can contain only assertions, and cannot nest other test cases or test groups.
```

content/languages/python/codewars-test/index.md
```diff
@@ -5,15 +5,59 @@ tags:
...
+
+
+### Assertions
+
+Assertions can be located only inside of a test case (a function decorated with `@it`). They must not be located directly under a test group (a function decorated with `@describe`), or in the top level of test.
+
+Every assertion generates separate log entry in the test output.
```
Steffan153, line 55:
```suggestion
Every assertion generates a separate log entry in the test output.
```

content/languages/python/codewars-test/index.md
```diff
@@ -59,113 +99,68 @@ The above produces an output similar to the following:
...
-Always put assertions inside `it` and not directly in `describe`.
-
-### Test Case
-
-`it` creates a test case containing assertion(s). Nesting will result in incorrect output.
 
 ## Failing Early
 
-Some of the functions below can accept a named argument `allow_raise=False`.
+Most of assertion functions can accept a named argument `allow_raise=False`.
```
Steffan153, line 105:
```suggestion
Most assertion functions can accept a named argument `allow_raise=False`.
```

content/languages/python/codewars-test/index.md
```diff
@@ -59,113 +99,68 @@ The above produces an output similar to the following:
...
-test.assert_approx_equals(actual, expected)
 test.assert_approx_equals(actual, expected, margin=1e-9, message=None, allow_raise=False)
-
-# default message: <actual> should be close to <expected> with absolute or relative margin of <margin>
 ```
 
-Checks if the actual value is close enough to the expected one, with a default relative or absolute value of `1e-9`.
-
-The comparison is done like this:
+Checks if the actual value is close enough to the expected one, with a default relative or absolute value of `margin`. The comparison is performed in following way:
```
Steffan153, line 141:
```suggestion
Checks if the actual value is close enough to the expected one, with a default relative or absolute value of `margin`. The comparison is performed like this:
```

content/languages/python/codewars-test/index.md
```diff
@@ -204,17 +197,17 @@ test.expect_error(msg, f, (OSError, KeyError)) # Pass
...
 
 ```python
-test.expect_no_error(message, function)
 test.expect_no_error(message, function, exception=BaseException)
 ```
 
-Checks this time that invoking `function` does **_not_** throw an exception of type `exception`.
+Checks whether invoked `function` does not throw an exception of a type specified by `exception`.
+
+The `exception` argument can specify a single type or a tuple of multiple exception types. Assertion is satisfied when no exception is raised, or when the raised exception is _not_ an instance of at least one of specified types. If during the execution of `function` an exception is raised that does _not_ match the parameter `exception`, it is silenced and the test is considered passed.
```
Steffan153, line 205:
```suggestion
The `exception` argument can specify a single type or tuple of multiple exception types. The assertion is satisfied when no exception is raised, or when the raised exception is _not_ an instance of at least one of specified types. If during the execution of `function` an exception is raised that does _not_ match the parameter `exception`, it is silenced and the test is considered passed.
```

content/languages/python/codewars-test/index.md
```diff
@@ -177,23 +172,21 @@ test.fail(message)
...
 test.expect_error(message, function, exception=Exception)
 ```
 
-Checks that invoking `function` throws an exception.
-If the argument `exception` is used, the raised exception must be an instance of that exception to consider the test as passed.
+Checks whether invoked `function` throws an exception. Raised exception must be an instance of a type specified with `exception` argument.
 
-- _Catching any exception:_ `Exception` is a catch-all type. So you can check _if a function throws anything_ doing the call without the `exception` argument.
-- _Catching specific exception(s):_ the `exception` argument can be a specific kind of exception class or even a tuple of multiple exception classes. The user throwing anyone of the specified exceptions will pass the test.
+The `exception` argument can specify a single type or a tuple of multiple exception types. Assertion is satisfied when the raised exception is an instance of at least one of specified types.
```
Steffan153, line 184:
```suggestion
The `exception` argument can specify a single type or a tuple of multiple exception types. The assertion is satisfied when the raised exception is an instance of at least one of the specified types.
```

content/languages/python/codewars-test/index.md
```diff
@@ -223,6 +216,19 @@ test.expect_no_error(msg, f, OSError)          # Pass
 
 ## Timeout Utility
 
+Runs the decorated function within the time limit.  
+
+`sec` is the amount of time allowed. It is expressed in seconds and can be given as an integer or float.  
+Generates a failed assertion when the function fails to complete in time, and its execution is terminated immediately.
+
+Decorated function is required to not throw any exception. If an error is raised during its execution, the test is considered failed and the error message becomes _"Should not throw any exceptions inside timeout: \<Exception()\>"_. This requirement is enforced by wrapping the inner function with `expect_no_error`, and as a side effect, one extra "test passed" entry is emitted for a collection of tests run inside a timeout wrapper.
```
Steffan153, line 224:
```suggestion
The decorated function is required to not throw any exceptions. If an error is raised during its execution, the test is considered failed and the error message becomes _"Should not throw any exceptions inside timeout: \<Exception()\>"_. This requirement is enforced by wrapping the inner function with `expect_no_error`, and as a side effect, one extra "test passed" entry is emitted for a collection of tests run inside a timeout wrapper.
```

content/languages/python/codewars-test/index.md
```diff
@@ -223,6 +216,19 @@ test.expect_no_error(msg, f, OSError)          # Pass
...
+
+:::warning
+Timed test should contain at least one assertion which verifies the result returned by the user solution. Otherwise, the test will be considered passed just if it happens to finish in time below the requested time limit, even if it would return incorrect answer.
```
Steffan153, line 227:
```suggestion
Timed tests should contain at least one assertion which verifies the result returned by the user solution. Otherwise, the test will be considered passed just if it happens to finish in time below the requested time limit, even if it would return an incorrect answer.
```

Steffan153 (changes requested):

content/languages/python/codewars-test/index.md
hobovsky: Actually, it should be `@codewars_test.describe` and `@codewars_test.it`. `@test.describe` and `@test.it` would work only if `codewars_test` would be aliased as `test` with `import codewars_test as test`, but it's not required to be done this way.

Steffan153: Well IMO it seems like it would be a pain doing `codewars_test.assert_equals` and etc. In all the _examples_ you always did like `@test.describe`/`@test.timeout`/etc, so you might as well be consistent.

Come to think of it, who knows who might do `from codewars_test import *` and eliminate all the `test.`s, but whatever...

content/languages/python/codewars-test/index.md
```diff
@@ -27,30 +27,40 @@ All functions, decorators, and assertions provided by the framework are defined
 The Python runner for versions prior to Python 3.8 does not contain the `codewars_test` module. The testing framework is imported implicitly and aliased as `test` and `Test`. This behavior is deprecated and for the Python 3.8 kata, the explicit import is required.
 :::
 
+:::note
+For simplicity, thorough the rest of this article it's assumed that `codewars_test` module is imported and aliased as `test` with following statement:
```
Steffan153, line 31:
```suggestion
For simplicity, through the rest of this article it's assumed that the `codewars_test` module is imported and aliased as `test` with the following statement:
```

content/languages/python/authoring/index.md
```diff
@@ -75,17 +75,25 @@ When any of these imports are missing, then, for compatibility reasons, it will
 
 If a user solution is expected to use functions or classes from the `preloaded` module, it's recommended to add required imports to the "initial solution" code snippet.
 
+:::note
+For simplicity, thorough the rest of this article it's assumed that `codewars_test` module is imported and aliased as `test` with following statement:
```
Steffan153, line 79:
```suggestion
For simplicity, through the rest of this article it's assumed that the `codewars_test` module is imported and aliased as `test` with the following statement:
```

Steffan153 (approved):

content/languages/python/codewars-test/index.md
hobovsky: I prefixed all mentions of decorators as `@test.describe` and `@test.it`. I left mentions of `describe` or `it` sections/blocks/groups without prefixes though, because they are not actual Python names, but a logical constructs.

Is it better now?

kazk (approved):

hobovsky:
Got two green marks so I am going to merge it, but if anyone, especially @Blind4Basics , has any remarks, then let me know or just create a PR with edits.

PR #212, opened 2020-12-30
JavaScript: Authoring kata

Steffan153:
Based on #189

Discussion: #214 
Deployment preview: [here](https://deploy-preview-212--reverent-edison-2864ea.netlify.app/languages/javascript/authoring/)


package.json
```diff
@@ -43,7 +43,7 @@
     "stylelint-config-prettier": "^8.0.2",
     "stylelint-config-standard": "^20.0.0",
     "stylelint-prettier": "^1.1.2",
-    "tailwindcss": "npm:@tailwindcss/postcss7-compat",
+    "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.0.2",
```
Steffan153, line 46:
```suggestion
    "tailwindcss": "npm:@tailwindcss/postcss7-compat",
```

Steffan153:
What about something about forbidding the use of something? You have to ban a ton of things for that, but it might be nice to mention something about it not being recommended, etc.

hobovsky:
> What about something about forbidding the use of something? You have to ban a ton of things for that, but it might be nice to mention something about it not being recommended, etc.

About this, I am thinking about creating a dedicated page which would explain why blocking modules and various restrictions on source code are generally not a good idea. Possibly, a set of language specific articles, for example "Python: blocking modules" would follow for authors who want to do this nonetheless despite of how bad idea this is.

In article "Create a kata in [some language]" I would either skip on this topic completely, or just hint on it or link to a dedicated article. There used to be a dedicated paragraph in the Python article I created, but I decided to remove it for now, and add a link to a dedicated article when it's done.

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,169 @@
...
+
+```javascript
+const chai = require('chai');
+const assert = chai.assert;
+```
+
+This is a convention used in many JavaScript kata, but it's not a requirement, and authors can choose to import the package in any way they find suitable for them. Also, the `should` or `expect` style can also be used from `chai`.
+:::
+
+JavaScript kata should use [Chai](https://www.chaijs.com/) to implement and execute tests. The most commonly used assertions include:
```
kazk, line 69: Chai is an assertion library. The test framework is [Mocha](https://mochajs.org/).

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,169 @@
...
+Some concepts don't always translate well to or from JavaScript. Because of this, some constructs should be avoided and some translations just shouldn't be done.
+- Avoid returning different data types depending on the situation (_"Return the result, or the string 'Error' if no result can be found..."_). JavaScript is dynamically typed, which is not the case for some other languages. Returning `null` might be appropriate in some situations, but throwing an error might be better in others.
+
+_TODO: Finish this_
+
+## Tests
+
+### Testing framework
+
+:::note
```
kazk, line 58: I think this note should be moved to after mentioning Chai.

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,169 @@
...
+- [`assert.strictEqual(actual, expected[, message])`](https://www.chaijs.com/api/assert/#method_strictequal) - Asserts that `actual === expected`.
+- [`assert.deepEqual(actual, expected[, message])`](https://www.chaijs.com/api/assert/#method_deepequal) - Asserts that two arrays or objects have the same values. `strictEqual` does not work since arrays and objects are reference types in JS.
+- [`assert.closeTo(actual, expected, delta[, message])`](https://www.chaijs.com/api/assert/#method_closeto) - Asserts that `abs(actual - expected) <= delta`. Should be used when comparing floating-point numbers.
+
+See the [Chai docs](https://www.chaijs.com/api/assert/) for all the methods.
+
+:::note
+When comparing arrays over 40 elements long, [the `truncateThreshold` option](https://www.chaijs.com/guide/styles/#configtruncatethreshold) should be set to `0` to make sure that the arrays will not be truncated in an error message. If the array is hundreds/thousands of elements long, do not do this as it will flood the console.
+```javascript
+chai.config.truncateThreshold = 0;
```
kazk, line 80: This is not rendered correctly for some reason. Try adding a line above the code block.

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,169 @@
...
+
+```javascript
+function random(lower, upper) {
+  return Math.floor(Math.random() * (upper - lower + 1) + lower);
+}
+
+random(1, 10); // inclusive random number between 1 and 10
+```
+
+However, to make this simpler, the [lodash](https://lodash.com) library includes some helpful functions to ease this:
```
kazk, line 146: We can also consider packaging commonly used test utilities. So you can do `const testUtils = require("@codewars/test-utils");` or something like that.

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,174 @@
...
+The most commonly used assertions include:
+
+- [`assert.strictEqual(actual, expected[, message])`](https://www.chaijs.com/api/assert/#method_strictequal) - Asserts that `actual === expected`.
+- [`assert.deepEqual(actual, expected[, message])`](https://www.chaijs.com/api/assert/#method_deepequal) - Asserts that two arrays or objects have the same values. `strictEqual` does not work since arrays and objects are reference types in JS.
+- [`assert.closeTo(actual, expected, delta[, message])`](https://www.chaijs.com/api/assert/#method_closeto) - Asserts that `abs(actual - expected) <= delta`. Should be used when comparing floating-point numbers.
+
+See the [Chai docs](https://www.chaijs.com/api/assert/) for all the methods.
+
+:::note
+When comparing arrays over 40 elements long, [the `truncateThreshold` option](https://www.chaijs.com/guide/styles/#configtruncatethreshold) should be set to `0` to make sure that the arrays will not be truncated in an error message. If the array is hundreds/thousands of elements long, do not do this as it will flood the console.
```
hobovsky, line 81: 40 seems to be somewhat specific here, and probably a configuration detail. Maybe just mention that `truncateThreshold` option exists, explain what it controls, and why authors should be careful?

```suggestion
The [`truncateThreshold` option](https://www.chaijs.com/guide/styles/#configtruncatethreshold) controls how arrays are stringified when `expected` and `actual` values are prepared for display in assertion messages. Setting it to 0 provides the most detailed feedback even for large arrays, but then assertion messages can flood the console and make the output difficult to read.
```

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+## Tasks and Requirements
+
+Some concepts don't always translate well to or from JavaScript. Because of this, some constructs should be avoided and some translations just shouldn't be done.
+- Avoid returning different data types depending on the situation (_"Return the result, or the string 'Error' if no result can be found..."_). JavaScript is dynamically typed, which is not the case for some other languages. Returning `null` might be appropriate in some situations, but throwing an error might be better in others.
+
+_TODO: Finish this_
+
+## Tests
```
hobovsky, line 54: I think organization of this part could be improved a bit, for example as described here: https://github.com/codewars/docs/issues/214#issuecomment-752796455

DonaldKellett (approved):

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,283 @@
...
+
+It's possible to put functions decorated with `it` in a loop and use them as a construct similar to parameterized test cases known from other testing frameworks, for example:
+
+```javascript
+describe("Generated test cases", () => {
+  const testCases = generateTestCases();
+  testCases.forEach(([msg, input, expected]) => {
+    it(msg, () => {
+      const actual = userSolution(input);
+      assert.strictEqual(actual, expected);
```
Steffan153, line 98: @hobovsky Now the problem here is, Chai hasn't been mentioned yet; and it's used here. Having it the other way around could confuse people with the `describe` and `it` though, so idk...

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,283 @@
+---
+kind: tutorial
+languages: [javascript]
+sidebar: "language:javascript"
+prev: /languages/javascript/codewars-test/
+---
+
+# JavaScript: creating and translating a kata
+
+This article is meant as a help for kata authors and translators who would like to create new content in JavaScript. It attempts to explain how to create and organize things in a way conforming to [authoring guidelines](/authoring/guidelines/), shows the most common pitfalls and how to avoid them.
```
DonaldKellett, line 10:
```suggestion
This article is intended for kata authors and translators who would like to create new content in JavaScript. It attempts to explain how to create and organize things in a way conforming to [authoring guidelines](/authoring/guidelines/), shows the most common pitfalls and how to avoid them.
```

DonaldKellett (changes requested):

content/languages/javascript/authoring.md
hobovsky: I don't think it's a big deal in this particular point, the way how it is now is ok-ish to me. If you are really concerned with the fact that `assert` is used without being introduced, then see if it helps when you move `Chai` part before `Mocha` part. As far as I see, the paragraph on assertions does not mention `describe` or `it` in any way?

DonaldKellett (approved):

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,285 @@
...
+
+:::
+
+### Test feedback
+
+Unlike Python or Ruby's test framework, Mocha handles combines multiple assertions in an `it` block into just one. For example, if all the assertions pass, then it will just show one `Test Passed`, and not a `Test Passed` for every single assertion. And if a test fails, it will not show all the other tests before/after it that passed/failed, it will only show the first failed test.
+
+### Generating random numbers
+
+JavaScript provides a simple [`Math.random`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random) function that returns a random floating-point number between `0` and `1`. You can use this to generate random numbers by multiplying it by another number, like `5`, and then flooring it, which returns a random number between 0 and **4**. Here's a simple utility to generate a random number between a lower and upper limit, inclusive:
```
kazk, line 146: > which returns a random number between 0 and **4**.

Is unstyled "0" intentional?

image

The strong **4** makes sense. 0 should be `0` like others for consistency?

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,285 @@
...
+However, to make this simpler, the [lodash](https://lodash.com) library includes some helpful functions to ease this:
+
+- [`_.random(lower, upper[, floating])`](https://lodash.com/docs#random) - Generates a random number between `lower` and `upper`, inclusively.
+- [`_.shuffle(array)`](https://lodash.com/docs#shuffle) - Returns shuffled `array`.
+- [`_.sample(array)`](https://lodash.com/docs#sample) - Returns a random element from `array`.
+- [`_.sampleSize(array, size)`](https://lodash.com/docs#sampleSize) - Returns a random number of elements up to `size` from `array`.
+
+### Additional packages
+
+The Codewars runner provides a set of preinstalled packages, which are available not only for users solving a kata, but can be also used by authors to build tests and generators of test cases. For example, `lodash` can be used to make it easier to generate random numbers/arrays.
```
kazk, line 165:
```suggestion
The Codewars runner provides a set of preinstalled packages, which are available not only for users solving a kata, but can be also used by authors to build tests and generators of test cases.
```
Redundant? Just talked about it right above.

kazk (approved):

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,285 @@
...
+
+```javascript
+chai.config.truncateThreshold = 0;
+```
+
+:::
+
+### Test feedback
+
+Unlike Python or Ruby's test framework, Mocha handles combines multiple assertions in an `it` block into just one. For example, if all the assertions pass, then it will just show one `Test Passed`, and not a `Test Passed` for every single assertion. And if a test fails, it will not show all the other tests before/after it that passed/failed, it will only show the first failed test.
```
hobovsky, line 142: It's not that assertions are "combined", it's rather that the test output panel presents the result of a test case, and not a result of an assertion. I am not even sure if I would mention it here, because (I think that) it's a normal behavior, expected from a sane testing framework, as opposed to log entry per assertion as it is for Python.

IMO this paragraph is useful for Python (and other languages which work in similar way), because there the test output is somewhat unusual, with one entry per assertion. I am not sure if I would put it for languages which behave in an expected way.

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,281 @@
...
+~~~
+```
+
+## Tasks and Requirements
+
+Some concepts don't always translate well to or from JavaScript. Because of this, some constructs should be avoided and some translations just shouldn't be done.
+- Avoid returning different data types depending on the situation (_"Return the result, or the string 'Error' if no result can be found..."_). JavaScript is dynamically typed, which is not the case for some other languages. Returning `null` might be appropriate in some situations, but throwing an error might be better in others.
+
+Some kata just should not be translated into JavaScript because it can be difficult to keep their initial idea:
+- Some kata might be meant for another language specifically. For example, a kata about learning to use Python's `itertools.takewhile` function to do something should not be translated to JavaScript.
```
hobovsky, line 53: I do not like the wording here, but it is a good point which could be turned into a general guideline: do not translate kata which are meant to be specific to some particular language, runtime, or a library, even if equivalents in other languages are available. For example, translating a kata like _"Learning Python: Classes"_ into JS might be not the best idea, even though both languages have classes.

Basically I would either reword this point (I know what you mean, and it can be just a language thing, but the mention of `itertools.takewhile` seems to be oddly specific to me and pops out somewhat suddenly here), or turn it into a general guideline and put it in [translating a kata](https://docs.codewars.com/authoring/guidelines/translation/) guidelines, for example in [this section](https://docs.codewars.com/authoring/guidelines/translation/#conformity-to-overall-kata-idea).

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,281 @@
...
+    // some assertions...
+  });
+});
+```
+
+`describe` blocks can contain `describe` or `it` blocks, but `it` blocks can only contain assertions.
+
+#### Dynamically generated test cases
+
+It's possible to put functions decorated with `it` in a loop and use them as a construct similar to parameterized test cases known from other testing frameworks, for example:
```
hobovsky, line 92: I think it's jsut a copy/paste mistake, but test functions in Mocha are not really decorated with anything.

```suggestion
It's possible to put calls to `it` in a loop and use them as a construct similar to parameterized test cases known from other testing frameworks, for example:
```

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,281 @@
...
+
+```javascript
+function random(lower, upper) {
+  return Math.floor(Math.random() * (upper - lower + 1) + lower);
+}
+
+random(1, 10); // inclusive random number between 1 and 10
+```
+
+However, to make this simpler, the [lodash](https://lodash.com) library includes some helpful functions to ease this:
```
hobovsky, line 152:
```suggestion
However, to make this simpler, the [lodash](https://lodash.com) library includes some helpful functions:
```

hobovsky: I find this sentence somewhat redundant, but please do not take my wording and grammar suggestions too seriously :)

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,281 @@
...
+    const arrCopy = [...arr];
+    // call user solution and ignore the result
+    userSolution(arrCopy);
+    // arrCopy should not be modified
+    assert.deepEqual(arrCopy, arr, 'Input array was modified');
+  });
+});
+
+describe('Random tests', () => {
+  function referenceSolution(arr) {
```
hobovsky, line 203: I believe the related paragraph is yet to come, but it would be worth to somehow stress in the example that it's important to put reference solutions in a limited scope.

```suggestion

  //reference solution enclosed in a local scope of `describe`
  function referenceSolution(arr) {
```

DonaldKellett (approved):

DonaldKellett (approved):

content/languages/javascript/authoring.md
```diff
@@ -0,0 +1,293 @@
...
+~~~
+```
+
+## Tasks and Requirements
+
+Some concepts don't always translate well to or from JavaScript. Because of this, some constructs should be avoided and some translations just shouldn't be done.
+- Avoid returning different data types depending on the situation (_"Return the result, or the string 'Error' if no result can be found..."_). JavaScript is dynamically typed, which is not the case for some other languages. Returning `null` might be appropriate in some situations, but throwing an error might be better in others.
+
+Some kata just should not be translated into JavaScript because it can be difficult to keep their initial idea:
+- Some kata might be meant for another language specifically. For example, a kata about `Python: Learning Classes` to do something should probably not be translated to JavaScript, even though JS has classes too.
```
hobovsky, line 53: Please someone verify the wording in this sentence, it sounds kinda off to me.

DonaldKellett, line 53:
```suggestion
- Some kata might be meant for another language specifically. For example, a kata about `Python: Learning Classes` should probably not be translated to JavaScript, even though JS has `class` syntax.
```

DonaldKellett: Technically speaking, JS only has objects and the prototype chain. `class` syntax should be viewed as syntactical sugar and not an actual JS construct.

DonaldKellett (changes requested):

hobovsky (approved):

content/languages/javascript/authoring.md
```diff
@@ -162,11 +162,21 @@ The Codewars runner provides a set of preinstalled packages, which are available
...
 ### Calling assertions
 
-_TODO_
+Chai provides a large set of useful [assertions](https://www.chaijs.com/api/assert/), but when used incorrectly, they can cause a series of problems:
+- Use of an assertion not suitable for the given case may lead to incorrect test results,
+- Incorrectly used assertions may produce confusing or unhelpful messages.
+
+To avoid the above problems, calls to assertion functions should respect the following rules:
+- The expected value should be calculated _before_ invoking an assertion. The `expected` parameter passed to the assertion should not be a function call expression, but a value calculated directly beforehand.
+- Appropriate assertion functions should be used for each given test. `assert.strictEqual` is not suitable in all situations. Use `assert.approximately` for floating-point comparisons, `assert.isOk` for tests on boolean values, and `assert.throws` to test error handling.
```
error256, line 177: Boolean things like `isOk` may be good for readability for a static test that is alone in its `it`. If we use custom messages or random tests, I think usual generic assertions should be more convenient for booleans.

hobovsky: I think I've seen users being puzzled by messages like _"Expected `True` to be `False`"_ or something like that, and IIRC it was emitted by `assert_equals(true, false)` i am just not sure in what language. They usually went "WTF how can `true` be false?"_. So a custom message would have to be used anyway.

Generally, custom messages should be more common. Test feedback of "Expected 5 to equal 3" is generally considered kinda WTFy in itself :)

error256: Yes, and that's because messages are supposed to be read in appropriate context. If `it` says  "should return False for n = 0" or at least "should work for n = 0", then any message is understandable, be it "Expected True to be False" or just "Test failed". With proper messages, it's all about readability and then there's not one true appropriate assertion.
I don't remember if custom messages in JS overwrite default messages of they add to them, but then custom messages make `isOk` appropriate for any comparison that's done manually anyway...

DonaldKellett (approved):

Issue #203, opened 2020-12-30
Create reference for "Kata Creator"


hobovsky:
Breakdown of #5 

Create a reference for kata creator/editor, potentially reusing the content from [old wiki](https://github.com/codewars/codewars.com/wiki/Kata-Creator).

Issue #202, opened 2020-12-30
Create "Kata trainer" reference


hobovsky:
Breakdown of #5 

Create a reference for kata trainer, potentially reusing the content from [old wiki](https://github.com/codewars/codewars.com/wiki/Kata-Trainer).

PR #221, opened 2021-01-02
Migrate ui references

hobovsky:
For the time being, just migrate existing things from old wiki:
- Kata Editor reference,
- Kata Trainer reference.

Discussion: #202 #203 
Preview: [here](https://deploy-preview-221--reverent-edison-2864ea.netlify.app/references/ui/)


hobovsky:
@kazk I think i will merge this as it was in old docs, an will try to improve it later. I just want to ask if you think it's in correct place.

content/references/gui/index.md
```diff
@@ -0,0 +1,11 @@
+---
+kind: reference
+sidebar: reference-gui
+next: /references/gui/kata-editor/
+---
+
+# Codewars GUI reference
```
kazk, line 7: I don't think we say GUI for Web apps because it's implied. Sounds awkward to me.

Maybe just `/references/ui/` and UI reference? Or don't even have this subdirectory and tag them with `ui`.

Importing it as is and improving later is fine.

hobovsky: > I don't think we say GUI for Web apps because it's implied. Sounds awkward to me.

Okay, it's probably language thing that it sounds ok to me. I will change GUI to UI.

> Or don't even have this subdirectory and tag them with `ui`.

I divided it into folders to keep future organization easier, because I expect to have many screenshots around here, so they will have their proper place instead of piling up in one dir.

hobovsky: I've replaced GUI with UI. If you have no more remarks, I will merge it.

Issue #207, opened 2020-12-30
Create a guide for authoring in C


hobovsky:
Create an article similar to https://docs.codewars.com/languages/python/authoring/ but for C.

Points needing particular attention:
- ~~memory management: show possible ways, common pitfalls, good practices~~ - moved to separate article
- assertions: Criterion assertions are macros with poor default messages
- includes
- bloated preloaded
- compilation warnings
- random utilities, `RAND_MAX`, `srand`

ggorlen:
There's a small walkthrough on improving criterion assertion messages [here](https://www.qualified.io/kb/languages/c/criterion) which may be useful.

hobovsky:
Thanks, can be useful when I start working on #232 👍

PR #228, opened 2021-01-09
Creating and translating C kata

hobovsky:
Discussion: #207 
Preview: [here](https://deploy-preview-228--reverent-edison-2864ea.netlify.app/languages/c/authoring/)

I will create separate PRs for:
- article on memory management pitfalls and practices,
- abridged Criterion reference


content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+- Default messages produced by assertion macros are confusing, so authors should provide custom messages for failed assertions.
+
+
+## Preloaded
+
+As C is a quite low level language, it often requires some boilerplate code to implment non-trivial tests, checks, and assertions. It can be tempting to put some code which would be common to sample tests and submission tests in the Preloaded snippet, but this approach sometimes proves to be problematic (see [here](/authoring/guidelines/preloaded/#accessibility-of-preloaded-code) why), and can cause some headaches for users who are interested in training on the kata locally, or checking how the user solution is called, etc. It's strongly discouraged to use preloaded code to make the code common for test snippets, if it would hide some information or implementation details interesting to the user.
```
Steffan153, line 138:
```suggestion
As C is a quite low-level language, it often requires some boilerplate code to implement non-trivial tests, checks, and assertions. It can be tempting to put some code that would be common to sample tests and submission tests in the Preloaded snippet, but this approach sometimes proves to be problematic (see [here](/authoring/guidelines/preloaded/#accessibility-of-preloaded-code) why), and can cause some headaches for users who are interested in training on the kata locally, or checking how the user solution is called, etc. It's strongly discouraged to use preloaded code to make the code common for test snippets if it would hide some information or implementation details interesting to the user. 
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+- Appropriate assertion functions should be used for each given test. `cr_assert_eq` is not suitable in all situations. Use `cr_assert_float_eq` for floating point comparisons, `cr_assert` for tests on boolean values, `cr_assert_str_*` to test strings and `cr_assert_arr_*` to test arrays.
+- Some additional attention should be paid to the order of parameters passed to assertion macros. It differs between various assertion libraries, and it happens to be quite often confused by authors, mixing up `actual` and `expected` in assertion messages. For the C testing framework, the order is `(actual, expected)`.
+- To avoid unexpected crashes in tests, it's recommended to perform some additional assertions before assuming that the answer returned by the user solution has some particular form, or size. For example, if the solution returns a pointer (possibly pointing to an array), an explicit assertion should be added to check whether the returned pointer is valid, and not, for example, `NULL`; size of the returned array, potentially reported by an output paramter, should be verified before accessing an element which could turn out to be located outside of its bounds.
```
Steffan153, line 132:
```suggestion
- To avoid unexpected crashes in tests, it's recommended to perform some additional assertions before assuming that the answer returned by the user solution has some particular form, or size. For example, if the solution returns a pointer (possibly pointing to an array), an explicit assertion should be added to check whether the returned pointer is valid, and not, for example, `NULL`; the size of the returned array, potentially reported by an output parameter, should be verified before accessing an element which could turn out to be located outside of its bounds.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+### Reference solution
+
+If the test suite happens to use a reference solution to calculate expected values (which [should be avoided](/authoring/guidelines/submission-tests/#reference-solution) when possible), or some kind of reference data like precalculated arrays, etc., it must not be possible for the user to call it, redefine, overwrite or directly access its contents. To prevent this, it should be defined as `static` in the tests implementation file.
+
+The reference solution or data ___must not___ be defined in the [Preloaded code](/authoring/guidelines/preloaded/).
+
+
+### Redeclaration of user solution
+
+Solution function should be redeclared in the file with submission tests. Such redeclaration prevents a compilation warning about implicitly declared functions, and additionally stops users from from tampering with the prototype of the solution function, for example to remove constness of parameters, or change types of parameters, etc.
```
Steffan153, line 114:
```suggestion
Solution function should be redeclared in the file with submission tests. Such redeclaration prevents a compilation warning about implicitly declared functions, and additionally stops users from tampering with the prototype of the solution function, for example, to remove constness of parameters, or change types of parameters, etc.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+
+### Input mutation
+
+General guidelines for submission tests contain a section related to [input mutation](/authoring/guidelines/submission-tests/#input-mutation) and how to prevent users from abusing it to work around kata requirements. Since C does not have reference semantics, it might appear that C kata are not affected by this problem, but it's not completely true. While data is passed to user solution by value, it indeed cannot be easily modified by user solution. However when data is passed indirectly, by a pointer or as an array, it can be modified _even when it's marked as `const`_. Constness of a function argument can be forcefuly cast away by a user and then they would be able to modify values passed as `const T*` or as elements of `const T[]`. It's usually not a problem in "real world" C programming, but on Codewars, users can take advantage on vulnerable test suites and modify their behavior this way. After calling a user solution, tests should not rely on the state of such values and they should consider them as potentially modified by a user.
```
Steffan153, line 118:
```suggestion
General guidelines for submission tests contain a section related to [input mutation](/authoring/guidelines/submission-tests/#input-mutation) and how to prevent users from abusing it to work around kata requirements. Since C does not have reference semantics, it might appear that C kata are not affected by this problem, but it's not completely true. While data is passed to the user solution by value, it indeed cannot be easily modified by the user solution. However, when data is passed indirectly, by a pointer, or as an array, it can be modified _even when it's marked as `const`_. Constness of a function argument can be forcefully cast away by a user and then they would be able to modify values passed as `const T*` or as elements of `const T[]`. It's usually not a problem in "real world" C programming, but on Codewars, users can take advantage of vulnerable test suites and modify their behavior this way. After calling a user solution, tests should not rely on the state of such values and they should consider them as potentially modified by a user.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+- Author might think that header files for the testing framework are included automatically by the code runner. It is not the case though, and test suites need to include `criterion/criterion.h` explicitly.
+
+### Compilation warnings
+
+Compiler options related to warnings used by C runner are somewhat strict and require some discipline to get the code to compile cleanly. `-Wall` and `-Wextra` cause that warnings can be numerous, and some of them are very pedantic. However, code of C kata should still compile in a clean way, without any warnings logged to the console. Even when a warnig does not cause any problem with tests, users get distracted by them and blame them for failing tests.
+
+
+### Memory management
+
+Unlike many modern, high level languages, C does not manage memory automatically. Manual memory management is a very vast and complex topic, with many possible ways of achieving the goal depending on a specific case, cave-ats, and pitfalls.
```
Steffan153, line 73:
```suggestion
Unlike many modern, high-level languages, C does not manage memory automatically. Manual memory management is a very vast and complex topic, with many possible ways of achieving the goal depending on a specific case, caveats, and pitfalls.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+- Author might think that header files for the testing framework are included automatically by the code runner. It is not the case though, and test suites need to include `criterion/criterion.h` explicitly.
+
+### Compilation warnings
+
+Compiler options related to warnings used by C runner are somewhat strict and require some discipline to get the code to compile cleanly. `-Wall` and `-Wextra` cause that warnings can be numerous, and some of them are very pedantic. However, code of C kata should still compile in a clean way, without any warnings logged to the console. Even when a warnig does not cause any problem with tests, users get distracted by them and blame them for failing tests.
```
Steffan153, line 68:
```suggestion
Compiler options related to warnings used by C runner are somewhat strict and require some discipline to get the code to compile cleanly. `-Wall` and `-Wextra` cause that warnings can be numerous, and some of them are very pedantic. However, code of C kata should still compile in a clean way, without any warnings logged to the console. Even when a warning does not cause any problem with tests, users get distracted by them and blame them for failing tests.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+- C is statically typed, so any task which depends on dynamic typing can be difficult to translate into C, and attempts of forcing a C kata to reflect dynamically typed interface can lead to ideas which enforce a really bad design.
+- There's just a few additional libraries available for C runner, so almost everything has to be implemented manually by the author or the user. Kata which take advantage of additional packages installed for other languages become much more difficult in C.
+
+## Coding
+
+### Code style
+
+Unlike for example Python or Java, there's no single guide for C code style, or even no set of naming conventions, which would be widely adopted and agreed upon by C programmers. There are traditional naming conventions using `snake_case`, there are Win32 API naming conventions using `PascalCase`, there are GNU guidelines, Microsoft guidelines, Google guidelines, and some of them contradictory to each other. Just use whatever set of guidelines you like, but when you do, use it consistently.
```
Steffan153, line 57:
```suggestion
Unlike for example Python or Java, there's no single guide for C code style, or even a set of naming conventions, which would be widely adopted and agreed upon by C programmers. There are traditional naming conventions using `snake_case`, there are Win32 API naming conventions using `PascalCase`, there are GNU guidelines, Microsoft guidelines, Google guidelines, and some of them contradictory to each other. Just use whatever set of guidelines you like, but when you do, use it consistently.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+~~~if-not:c
+
+...text not visible in C description...
+
+~~~
+```
+
+## Tasks and Requirements
+
+Some concepts don't always translate well to or from C. Because of this, some constructs should be avoided and some translations just shouldn't be done. Some high level languages, like Pyton or Javascript, reside on the exactly opposite end of the spectrum than C, and translating kata between them and C can result in signiicant differences in difficulty, requirements, and design of the solution.
```
Steffan153, line 48:
```suggestion
Some concepts don't always translate well to or from C. Because of this, some constructs should be avoided and some translations just shouldn't be done. Some high-level languages, like Python or JavasScript, reside on the exact opposite end of the spectrum than C, and translating kata between them and C can result in significant differences in difficulty, requirements, and design of the solution.
```

content/languages/c/criterion.md
```diff
@@ -9,6 +9,8 @@ tags:
 
 # Criterion
 
+Criterion testing framework is available in its [Github repository](https://github.com/Snaipe/Criterion). Reference for Criterion API can be found here: https://criterion.readthedocs.io/en/master/intro.html .
```
Steffan153, line 12:
```suggestion
The Criterion testing framework is available in its [Github repository](https://github.com/Snaipe/Criterion). Reference for the Criterion API can be found here: [https://criterion.readthedocs.io/en/master/intro.html](https://criterion.readthedocs.io/en/master/intro.html).
```
(the explicit link text is to remove the space before period)

Steffan153 (changes requested):

content/languages/c/authoring/index.md
hobovsky: I can't see any change highlighted here, is there any suggestion hiding here I cannot see?

Steffan153: There's a suggestion, the line is pretty big though.
image

Steffan153: If you compare side-by-side, there are some differences. Not sure why GH is acting up.

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Arrays and strings
+
+- malloc in solution and free in tests
+- pass in a preallocated buffer (use size hints if possible)
+- two functions: get size, allocate in tests, run solution
+- two functions: solution with allocation, deallocation. Bookkeeping information managed by user or passed as additional `void*`
+- one function: accept buffer+size, return retsult or error and required size
+- avoid string constants, use named symbols
+- reporting size:output param, structure, sentinel teminators
```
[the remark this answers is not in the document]
hobovsky: It's just a checklist for another article I am going to work on soon, I will create a dedicated PR for it.

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+
+...text not visible in C description...
+
+~~~
+```
+
+## Tasks and Requirements
+
+Some concepts don't always translate well to or from C. Because of this, some constructs should be avoided and some translations just shouldn't be done. Some high-level languages, like Python or JavasScript, reside on the exact opposite end of the spectrum than C, and translating kata between them and C can result in significant differences in difficulty, requirements, and design of the solution.
+- C is a language of much lower level than many other popular languages available on Codewars. For this reason, many kata, even if their task can be translated to C directly, can turn out much harder in C than in original language. There's many kata which were originally created as very easy and beginner friendly (for example 8 kyu). But after translating into C, and adding aspects like memory management, or two dimensional C arrays, etc. they are not so easy anymore, and newbies complain that kata ranked 8 kyu is too difficult for them while it should be an entry level task.
```
Steffan153, line 49:
```suggestion
- C is a much lower-level language than many other popular languages available on Codewars. For this reason, many kata, even if their task can be translated to C directly, can turn out much harder in C than in the original language. There are many kata that were originally created as very easy and beginner-friendly (for example 8 kyu). But after translating into C, and adding aspects like memory management, or two dimensional C arrays, etc. they are not so easy anymore, and newbies complain that kata ranked 8 kyu is too difficult for them while it should be an entry-level task.  
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+- C is a language of much lower level than many other popular languages available on Codewars. For this reason, many kata, even if their task can be translated to C directly, can turn out much harder in C than in original language. There's many kata which were originally created as very easy and beginner friendly (for example 8 kyu). But after translating into C, and adding aspects like memory management, or two dimensional C arrays, etc. they are not so easy anymore, and newbies complain that kata ranked 8 kyu is too difficult for them while it should be an entry level task.  
+- C is statically typed, so any task which depends on dynamic typing can be difficult to translate into C, and attempts of forcing a C kata to reflect dynamically typed interface can lead to ideas which enforce a really bad design.
```
Steffan153, line 50:
```suggestion
- C is statically typed, so any task that depends on dynamic typing can be difficult to translate into C, and attempts of forcing a C kata to reflect a dynamically typed interface can lead to ideas that enforce a really bad design.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+- There's just a few additional libraries available for C runner, so almost everything has to be implemented manually by the author or the user. Kata which take advantage of additional packages installed for other languages become much more difficult in C.
```
Steffan153, line 51:
```suggestion
- There are just a few additional libraries available for C runner, so almost everything has to be implemented manually by the author or the user. Kata that take advantage of additional packages installed for other languages become much more difficult in C.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+
+## Coding
+
+### Code style
+
+Unlike for example Python or Java, there's no single guide for C code style, or even a set of naming conventions, which would be widely adopted and agreed upon by C programmers. There are traditional naming conventions using `snake_case`, there are Win32 API naming conventions using `PascalCase`, there are GNU guidelines, Microsoft guidelines, Google guidelines, and some of them contradictory to each other. Just use whatever set of guidelines you like, but when you do, use it consistently.
```
Steffan153, line 57:
```suggestion
Unlike for example Python or Java, there's no single guide for C code style, or even a set of naming conventions, which would be widely adopted and agreed upon by C programmers. Traditional naming conventions are using `snake_case`, Win32 API naming conventions are using `PascalCase`, there are GNU guidelines, Microsoft guidelines, Google guidelines, and some of them contradictory to each other. Just use whatever set of guidelines you like, but when you do, use it consistently.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+
+### Header files
+
+Not as much of a problem for C as it is for C++, but still, C authors often forget to include required header files, or just leave them out deliberately because "it works" even when some of the files are not included. It happens mostly due to following reasons:
```
Steffan153, line 61:
```suggestion
Not as much of a problem for C as it is for C++, but still, C authors often forget to include required header files, or just leave them out deliberately because "it works" even when some of the files are not included. It happens mostly due to the following reasons:
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+- Compiler provides implicit declaration of a function, when it's encountered in the code and was not declared. However, this behavior is not standard and is now deprecated. You need to explicitly include header files for library functions you use, or declare them in some other way.
```
Steffan153, line 62:
```suggestion
- The compiler provides an implicit declaration of a function, when it's encountered in the code and was not declared. However, this behavior is not standard and is now deprecated. You need to explicitly include header files for library functions you use or declare them in some other way.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+- Some header files include other header files indirectly, for example, file `foo.h` contains line `#include <bar.h>`, which might appear to make the explicit include for `bar.h` unnecessary. It's not true though, because the file `foo.h` might change one day, or might depend on some compiler settings or command line options, and after some changes to the cnfiguration of the C runner, the `bar.h` might be not included there anymore. That's why every file (i.e. code snippet) of a kata should explicitly include all required header files declaring functions used in it.
```
Steffan153, line 63:
```suggestion
- Some header files include other header files indirectly, for example, file `foo.h` contains line `#include <bar.h>`, which might appear to make the explicit include for `bar.h` unnecessary. It's not true though, because the file `foo.h` might change one day, or might depend on some compiler settings or command line options, and after some changes to the configuration of the C runner, the `bar.h` might be not included there anymore. That's why every file (i.e. code snippet) of a kata should explicitly include all required header files declaring functions used in it.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+- Author might think that header files for the testing framework are included automatically by the code runner. It is not the case though, and test suites need to include `criterion/criterion.h` explicitly.
+
+### Compilation warnings
+
+Compiler options related to warnings used by C runner are somewhat strict and require some discipline to get the code to compile cleanly. `-Wall` and `-Wextra` cause that warnings can be numerous, and some of them are very pedantic. However, code of C kata should still compile in a clean way, without any warnings logged to the console. Even when a warning does not cause any problem with tests, users get distracted by them and blame them for failing tests.
```
Steffan153, line 68:
```suggestion
Compiler options related to warnings used by C runner are somewhat strict and require some discipline to get the code to compile cleanly. `-Wall` and `-Wextra` cause that warnings can be numerous, and some of them are very pedantic. However, code of C kata should still compile cleanly, without any warnings logged to the console. Even when a warning does not cause any problem with tests, users get distracted by them and blame them for failing tests.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+
+
+### Memory management
+
+Unlike many modern, high-level languages, C does not manage memory automatically. Manual memory management is a very vast and complex topic, with many possible ways of achieving the goal depending on a specific case, caveats, and pitfalls.
+
+Whenever a kata needs to return a string or an array, C authors tend to use the naive technique of allocating the memory in the solution function, and freeing it in the test suite. This approach mimics the behavior known from other languages where returning an array or object from inside of user solution is perfectly valid, but it's hardly ever a valid way of working with unmanaged memory.
+
+Possible ways of handling memory management are described in the [Memory Management in C kata](/languages/c/authoring/memory-management-techniques/) article. But whichever approach is chosen, even the most obvious one, it should be described either in the kata description (preferrably in in a C-specific paragraph), or in the initial solution stub as a comment, and in sample tests as an example of a call to the solution.
```
Steffan153, line 77:
```suggestion
Possible ways of handling memory management are described in the [Memory Management in C kata](/languages/c/authoring/memory-management-techniques/) article. But whichever approach is chosen, even the most obvious one, it should be described either in the kata description (preferably in a C-specific paragraph), or in the initial solution stub as a comment, and in sample tests as an example of a call to the solution.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+
+## Tests
+
+### Testing framework
+
+C kata use the [Criterion testing framework](/languages/c/criterion/) to implement and execute tests. Read its reference page to find out how to structure tests into groups and test cases, what assertions are available, etc.
+
+Criterion supports many features which can be very helpful, but (unfortunately) are not commonly used by C authors. It allows for parametrized tests, setting up additional data, test fixture setup and teardown, custom descriptions, etc.
```
Steffan153, line 85:
```suggestion
Criterion supports many features that can be very helpful, but (unfortunately) are not commonly used by C authors. It allows for parameterized tests, setting up additional data, test fixture setup, teardown, custom descriptions, etc. 
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+
+### Testing framework
+
+C kata use the [Criterion testing framework](/languages/c/criterion/) to implement and execute tests. Read its reference page to find out how to structure tests into groups and test cases, what assertions are available, etc.
+
+Criterion supports many features which can be very helpful, but (unfortunately) are not commonly used by C authors. It allows for parametrized tests, setting up additional data, test fixture setup and teardown, custom descriptions, etc. 
+
+### Test feedback
+
+You should notice that the report hooks used by Codewars test runner produce one test output entry per assertion, so the test output panel can get very noisy.
```
Steffan153, line 89:
```suggestion
You should notice that the report hooks used by the Codewars test runner produce one test output entry per assertion, so the test output panel can get very noisy.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+
+
+### Random utilities
+
+Unlike some other languages, C does not provide too many means of generating random numbers which could be used to build random tests. `stdlib.h` header provides the `rand` function which, while being quite simple, satisfies majority of needs, but sometimes can be tricky to be used correctly.
```
Steffan153, line 94:
```suggestion
Unlike some other languages, C does not provide too many means of generating random numbers which could be used to build random tests. `stdlib.h` header provides the `rand` function which, while being quite simple, satisfies the majority of needs, but sometimes can be tricky to be used correctly.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+
+Before `rand` is called for the first time, it must be seeded with `srand`. A call to `srand` should be performed only once, in the setup phase of the random tests. `srand` usually uses current time as a seed, so authors need to include `time.h` before using `time` function.
```
Steffan153, line 96:
```suggestion
Before `rand` is called for the first time, it must be seeded with `srand`. A call to `srand` should be performed only once, in the setup phase of the random tests. `srand` usually uses the current time as a seed, so authors need to include `time.h` before using the `time` function.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+
+`rand` can return integers only up to `RAND_MAX`. There's no standard-compliant way to generate random values of types `unsigned int`, `long`, or `double`. Authors which would like to generate random values out of domain of `rand` have to craft them manually. _(TODO: create article with snippets with RNGs for types other than `int`)_
```
Steffan153, line 98:
```suggestion
`rand` can return integers only up to `RAND_MAX`. There's no standard-compliant way to generate random values of types `unsigned int`, `long`, or `double`. Authors who would like to generate random values out of the domain of `rand` have to craft them manually. _(TODO: create article with snippets with RNGs for types other than `int`)_
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+
+Additionally, value of `RAND_MAX` might differ on different platforms, or even change. For current Codewars setup it's `2^32-1`, but there are some common platforms with `RAND_MAX` being as small as `2^16-1`. This makes the code using `rand` even less portable, and while portability might not be a big concern for Codewars kata, it could turn out to be an issue for users trying to reproduce random tests locally.
```
Steffan153, line 100:
```suggestion
Additionally, the value of `RAND_MAX` might differ on different platforms, or even change. For the current Codewars setup it's `2^32-1`, but there are some common platforms with `RAND_MAX` being as small as `2^16-1`. This makes the code using `rand` even less portable, and while portability might not be a big concern for Codewars kata, it could turn out to be an issue for users trying to reproduce random tests locally.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+
+An alternative to `rand` could be using random devices, like `/dev/urandom`. This way of generating random numbers could partially alleviate the issue of the `rand` being capped at `RAND_MAX`, but also could inflate amount of the bolierplate code and could cause additional problems with portability.
```
Steffan153, line 102:
```suggestion
An alternative to `rand` could be using random devices, like `/dev/urandom`. This way of generating random numbers could partially alleviate the issue of the `rand` being capped at `RAND_MAX`, but also could inflate the amount of the boilerplate code and could cause additional problems with portability.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,278 @@
...
+//reference solution defined as static
+static void square_every_item_ref(double items[], int size)
+{
+    for(int i = 0; i<size; ++i)
+    {
+      items[i] *= items[i];
+    }
+}
+
+//custom comparer for floating point values
```
Steffan153, line 167:
```suggestion
//custom comparer for floating-point values
```

Steffan153 (changes requested):
welp

Steffan153 (approved):

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+
+...text not visible in C description...
+
+~~~
+```
+
+## Tasks and Requirements
+
+Some concepts don't always translate well to or from C. Because of this, some constructs should be avoided and some translations just shouldn't be done. Some high-level languages, like Python or JavasScript, reside on the exact opposite end of the spectrum than C, and translating kata between them and C can result in significant differences in difficulty, requirements, and design of the solution.
+- C is a language of much lower level than many other popular languages available on Codewars. For this reason, many kata, even if their task can be translated to C directly, can turn out much harder in C than in original language. There's many kata which were originally created as very easy and beginner friendly (for example 8 kyu). But after translating into C, and adding aspects like memory management, or two dimensional C arrays, etc. they are not so easy anymore, and newbies complain that kata ranked 8 kyu is too difficult for them while it should be an entry level task.
```
Steffan153, line 49: ...how did this get reverted?

hobovsky: Not all, only on this paragraph, I think. I probably made some mistake when resolving a merge conflict. Could you re-suggest them? I think other paragraphs were merged correctly?

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+- C is statically typed, so any task which depends on dynamic typing can be difficult to translate into C, and attempts of forcing a C kata to reflect dynamically typed interface can lead to ideas which enforce a really bad design.
```
Steffan153, line 50:
```suggestion
- C is statically typed, so any task that depends on dynamic typing can be difficult to translate into C, and attempts of forcing a C kata to reflect a dynamically typed interface can lead to ideas that enforce a really bad design.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+- There's just a few additional libraries available for C runner, so almost everything has to be implemented manually by the author or the user. Kata which take advantage of additional packages installed for other languages become much more difficult in C.
```
Steffan153, line 51:
```suggestion
- There are just a few additional libraries available for C runner, so almost everything has to be implemented manually by the author or the user. Kata that take advantage of additional packages installed for other languages become much more difficult in C.
```

content/languages/c/authoring/index.md
Steffan153:
```suggestion
- C is much lower-level than many other popular languages available on Codewars. For this reason, many kata, even if their task can be translated to C directly, can turn out much harder in C than in the original language. There are many kata that were originally created as very easy and beginner-friendly (for example 8 kyu). But after translating into C, and adding aspects like memory management, or two dimensional C arrays, etc. they are not so easy anymore, and newbies complain that kata ranked 8 kyu is too difficult for them while it should be an entry-level task.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+~~~if-not:c
+
+...text not visible in C description...
+
+~~~
+```
+
+## Tasks and Requirements
+
+Some concepts don't always translate well to or from C. Because of this, some constructs should be avoided and some translations just shouldn't be done. Some high-level languages, like Python or JavasScript, reside on the exact opposite end of the spectrum than C, and translating kata between them and C can result in significant differences in difficulty, requirements, and design of the solution.
```
Steffan153, line 48:
```suggestion
Some concepts don't always translate well to or from C. Because of this, some constructs should be avoided and some translations just shouldn't be done. Some high-level languages, like Python or JavaScript, reside on the exact opposite end of the spectrum than C, and translating kata between them and C can result in significant differences in difficulty, requirements, and design of the solution.
```
This actually did get merged properly, I just made a mistake...

Steffan153 (changes requested):

Steffan153 (approved):

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
+---
+kind: tutorial
+languages: [c]
+sidebar: "language:c"
+---
+
+# C: creating and translating a kata
+
+This article is meant as help for kata authors and translators who would like to create new content in C. It attempts to explain how to create and organize things in a way conforming to [authoring guidelines](/authoring/guidelines/), shows the most common pitfalls and how to avoid them.
```
DonaldKellett, line 9:
```suggestion
This article is meant for kata authors and translators who would like to create new content in C. It attempts to explain how to create and organize things in a way conforming to [authoring guidelines](/authoring/guidelines/), shows the most common pitfalls and how to avoid them.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+- C is much lower-level than many other popular languages available on Codewars. For this reason, many kata, even if their task can be translated to C directly, can turn out much harder in C than in the original language. There are many kata that were originally created as very easy and beginner-friendly (for example 8 kyu). But after translating into C, and adding aspects like memory management, or two dimensional C arrays, etc. they are not so easy anymore, and newbies complain that kata ranked 8 kyu is too difficult for them while it should be an entry-level task.
+- C is statically typed, so any task that depends on dynamic typing can be difficult to translate into C, and attempts of forcing a C kata to reflect a dynamically typed interface can lead to ideas that enforce a really bad design.
+- There are just a few additional libraries available for C runner, so almost everything has to be implemented manually by the author or the user. Kata that take advantage of additional packages installed for other languages become much more difficult in C.
```
DonaldKellett, line 51:
```suggestion
- There are just a few additional libraries available for the C runner, so almost everything has to be implemented manually by the author or the user. Kata that take advantage of additional packages installed for other languages become much more difficult in C.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Coding
+
+### Code style
+
+Unlike for example Python or Java, there's no single guide for C code style, or even a set of naming conventions, which would be widely adopted and agreed upon by C programmers. Traditional naming conventions are using `snake_case`, Win32 API naming conventions are using `PascalCase`, there are GNU guidelines, Microsoft guidelines, Google guidelines, and some of them contradictory to each other. Just use whatever set of guidelines you like, but when you do, use it consistently.
```
DonaldKellett, line 57:
```suggestion
Unlike for example Python or Java, there's no single guide for C code style, or even a set of naming conventions, which would be widely adopted and agreed upon by C programmers. Traditional naming conventions are using `snake_case`, Win32 API naming conventions are using `PascalCase`, there are GNU guidelines, Microsoft guidelines, Google guidelines, and some of them contradict each other. Just use whatever set of guidelines you like, but when you do, use it consistently.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+- The compiler provides an implicit declaration of a function, when it's encountered in the code and was not declared. However, this behavior is not standard and is now deprecated. You need to explicitly include header files for library functions you use or declare them in some other way.
+- Some header files include other header files indirectly, for example, file `foo.h` contains line `#include <bar.h>`, which might appear to make the explicit include for `bar.h` unnecessary. It's not true though, because the file `foo.h` might change one day, or might depend on some compiler settings or command line options, and after some changes to the configuration of the C runner, the `bar.h` might be not included there anymore. That's why every file (i.e. code snippet) of a kata should explicitly include all required header files declaring functions used in it.
+- Author might think that header files for the testing framework are included automatically by the code runner. It is not the case though, and test suites need to include `criterion/criterion.h` explicitly.
```
DonaldKellett, line 64:
```suggestion
- The author might think that header files for the testing framework are included automatically by the code runner. That is not the case though, and test suites need to include `criterion/criterion.h` explicitly.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Compilation warnings
+
+Compiler options related to warnings used by C runner are somewhat strict and require some discipline to get the code to compile cleanly. `-Wall` and `-Wextra` cause that warnings can be numerous, and some of them are very pedantic. However, code of C kata should still compile cleanly, without any warnings logged to the console. Even when a warning does not cause any problem with tests, users get distracted by them and blame them for failing tests.
```
DonaldKellett, line 68:
```suggestion
Compiler options related to warnings used by the C runner are somewhat strict and require some discipline to get the code to compile cleanly. `-Wall` and `-Wextra` may cause numerous warnings and some of them are very pedantic. However, code of C kata should still compile cleanly, without any warnings logged to the console. Even when a warning does not cause any problem with tests, users get distracted by them and blame them for failing tests.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+
+
+### Memory management
+
+Unlike many modern, high-level languages, C does not manage memory automatically. Manual memory management is a very vast and complex topic, with many possible ways of achieving the goal depending on a specific case, caveats, and pitfalls.
+
+Whenever a kata needs to return a string or an array, C authors tend to use the naive technique of allocating the memory in the solution function, and freeing it in the test suite. This approach mimics the behavior known from other languages where returning an array or object from inside of user solution is perfectly valid, but it's hardly ever a valid way of working with unmanaged memory.
```
DonaldKellett, line 75:
```suggestion
Whenever a kata needs to return a string or an array, C authors tend to use the naive technique of allocating the memory in the solution function, and freeing it in the test suite. This approach mimics the behavior known from other languages where returning an array or object from inside of the user solution is perfectly valid, but it's hardly ever a valid way of working with unmanaged memory.
```

DonaldKellett: I'm always been torn over this - correct English usage should be "the user's solution", but the "user solution" seems to be a somewhat accepted Codewars terminology. Which one should we stick to?

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+
+One of the consequences of non-managed memory is that it's strongly recommended to avoid requirement of returning string constants from C solution, especially when translating kata from other languages. Returning a string in other languages is not a problem, but in C it always rises questions of who and how should allocate it. Consider replacing the string with some simpler data type (eventually aliased with a `typedef`), and/or provide some symbolic constants for available values. For exmple, if for Python version the requirement would be to _Return string `'You won!'` if you won the game, `'You lost :('` if you were defeated, and `'It's a draw'` if there is no winner._, C should preferably use constants of `WON`, `LOST` and `DRAW`. If author decides to keep raw C-strings as elements of the kata interface, they should clearly specify required alloction scheme.
```
DonaldKellett, line 77:
```suggestion
One of the consequences of unmanaged memory is that it's strongly recommended against returning string constants from C functions, especially when translating kata from other languages. Returning a string in other languages is not a problem, but in C it always raises questions of who should allocate it and how it should be allocated. Consider replacing the string with some simpler data type (eventually aliased with a `typedef`), and/or provide some symbolic constants for available values. For example, if the requirement for the Python version is to _Return the string `'You won!'` if you won the game, `'You lost :('` if you were defeated, and `'It's a draw'` if there is no winner._, C should preferably use the constants `WON`, `LOST` and `DRAW` defined in an `enum`. If the author decides to keep raw C-strings as elements of the kata interface, they should clearly specify the required allocation scheme.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+
+Possible ways of handling memory management are described in the [Memory Management in C kata](/languages/c/authoring/memory-management-techniques/) article. But whichever approach is chosen, even the most obvious one, it should be described either in the kata description (preferrably in in a C-specific paragraph), or in the initial solution stub as a comment, and in sample tests as an example of a call to the solution.
```
DonaldKellett, line 79:
```suggestion
Possible ways of handling memory management are described in the [Memory Management in C kata](/languages/c/authoring/memory-management-techniques/) article. But whichever approach is chosen, even the most obvious one, it should be described either in the kata description (preferably in in a C-specific paragraph), or in the initial solution stub as a comment, and in sample tests as an example of a call to the solution.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+### Reference solution
+
+If the test suite happens to use a reference solution to calculate expected values (which [should be avoided](/authoring/guidelines/submission-tests/#reference-solution) when possible), or some kind of reference data like precalculated arrays, etc., it must not be possible for the user to call it, redefine, overwrite or directly access its contents. To prevent this, it should be defined as `static` in the tests implementation file.
+
+The reference solution or data ___must not___ be defined in the [Preloaded code](/authoring/guidelines/preloaded/).
+
+
+### Redeclaration of user solution
+
+Solution function should be redeclared in the file with submission tests. Such redeclaration prevents a compilation warning about implicitly declared functions, and additionally stops users from tampering with the prototype of the solution function, for example, to remove constness of parameters, or change types of parameters, etc.
```
DonaldKellett, line 116:
```suggestion
The solution function should be re-declared in the file with submission tests. Such re-declaration prevents a compilation warning about implicitly declared functions, and additionally stops users from tampering with the prototype of the solution function, for example, to remove const-ness of parameters, or change types of parameters, etc.
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+
+
+### Calling assertions
+
+Criterion testing framework provides a set of useful [assertions](/languages/c/criterion/#assertions), but when used incorrectly, they can cause a series of problems:
```
DonaldKellett, line 125:
```suggestion
Criterion provides a set of useful [assertions](/languages/c/criterion/#assertions), but when used incorrectly, they can cause a series of problems:
```

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+- To avoid unexpected crashes in tests, it's recommended to perform some additional assertions before assuming that the answer returned by the user solution has some particular form, or size. For example, if the solution returns a pointer (possibly pointing to an array), an explicit assertion should be added to check whether the returned pointer is valid, and not, for example, `NULL`; the size of the returned array, potentially reported by an output parameter, should be verified before accessing an element which could turn out to be located outside of its bounds.
+- Default messages produced by assertion macros are confusing, so authors should provide custom messages for failed assertions.
+
+
+### Testability
+
+In C, not everything can be easily tested. It's not possible to reliably verify the size or bounds of returned buffer, or a correctness of returned pointer. It's difficult to test for conditions which result in a crash or undefined behavior. It cannot be reliably verified whether there's no memory leaks and if all allocated memory were correctly released. Sometimes the only way is to skip some checks or crash the tests.
```
DonaldKellett, line 140:
```suggestion
In C, not everything can be easily tested. It's not possible to reliably verify the size or bounds of a returned buffer, or the validity of a returned pointer. It's difficult to test for conditions which result in a crash or undefined behavior. It cannot be reliably verified whether there's no memory leaks and if all allocated memory were correctly released. Sometimes the only way is to skip some checks or crash the tests.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -0,0 +1,24 @@
...
+
+_TBD_
+
+## Arrays and strings
+
+- malloc in solution and free in tests
+- pass in a preallocated buffer (use size hints if possible)
+- two functions: get size, allocate in tests, run solution
+- two functions: solution with allocation, deallocation. Bookkeeping information managed by user or passed as additional `void*`
+- one function: accept buffer+size, return retsult or error and required size
```
DonaldKellett, line 17:
```suggestion
- one function: accept buffer+size, return result or error and required size
```

content/languages/c/criterion.md
```diff
@@ -30,6 +32,15 @@ Test(add_test, should_add_integers, .description = "should add integers") {
 }
 ```
 
+## Assertions
+
+_TBD_
+
+[Assertions reference](https://criterion.readthedocs.io/en/master/assert.html) in Criterion documentation.
```
DonaldKellett, line 39:
```suggestion
[Assertions reference](https://criterion.readthedocs.io/en/master/assert.html) can be found in the Criterion documentation.
```

DonaldKellett (changes requested):

content/languages/c/authoring/index.md
hobovsky: I've deliberately not included the _"defined in an `enum`"_ part here, and leave authors more freedom on this subject. It's true that an enum would be the best thing style-wise, but in a kata, if they want, they can use `int`, `char`, or even `const char *const` with full message content defined in preloaded, as long as it does not leak to the user to make them worry about memory allocation.

I will remove _"defined in an `enum`"_ part and merge the wording suggestions, but if you think it must stay, then let me know.

content/languages/c/authoring/index.md
hobovsky: I think I'd prefer the correct one more than "ours" :)
```suggestion
Whenever a kata needs to return a string or an array, C authors tend to use the naive technique of allocating the memory in the solution function, and freeing it in the test suite. This approach mimics the behavior known from other languages where returning an array or object from inside of the user's solution is perfectly valid, but it's hardly ever a valid way of working with unmanaged memory.
```

content/languages/c/authoring/memory-management-techniques.md
hobovsky: This file is going to have a separate PR, there's no need to fix typos here.

content/languages/c/criterion.md
hobovsky: This file will have its own PR too, but I will merge this one since the file is already there, potentially known to the users.

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+
+One of the consequences of unmanaged memory is that it's strongly recommended against returning string constants from C functions, especially when translating kata from other languages. Returning a string in other languages is not a problem, but in C it always raises questions of who should allocate it and how it should be allocated. Consider replacing the string with some simpler data type (eventually aliased with a `typedef`), and/or provide some symbolic constants for available values. For example, if the requirement for the Python version is to _Return the string `'You won!'` if you won the game, `'You lost :('` if you were defeated, and `'It's a draw'` if there is no winner._, C should preferably use the constants `WON`, `LOST` and `DRAW`. If the author decides to keep raw C-strings as elements of the kata interface, they should clearly specify the required allocation scheme.
```
DonaldKellett, line 77: Is there a reason why plain constants are preferred over `enum`s for a user-defined set of values (referring to "C should preferably use the constants `WON`, `LOST` and `DRAW`")? IMO the latter better communicates the intent that these user-defined values are related to each other and constitute a "type" of their own.

DonaldKellett (commented):
See the part regarding constants versus `enum`s.

content/languages/c/authoring/index.md
hobovsky: Have you seen my comment [here](https://github.com/codewars/docs/pull/228#discussion_r554881117)?

Of course, from the design point of view, you are completely right. However, I am somewhat concerned with how such guideline would be received by Codewars authors. I can imagine a situation when a Python kata, for example [Red Knight](https://www.codewars.com/kata/5fc4349ddb878a0017838d0f/), has a return type of `(string, int)` and now C translator would like to have return values of `{ "White", 10 }` or `{"Black", 42}` (i.e. to keep the requirement of having a string, what was actually the case), instead of introducing an enum. I do not know why they would want to, but I've already seen various explanations, starting from "because consistency" to "`char` abbreviation will be OK I like it more`" to "I don't like enums". My goal for this particular guideline is to remove the ambuiguity of the allocation type which should be used, and not enforce some particular coding style. As much as I personally would prefer, use, and suggest an enum here, I do not want to strictly force authors to do so and I want to give them a choice.

As I said, I do not see any problem with putting your part on enums here other than authors feeling forced. You can re-suggest it, and I will merge it.

DonaldKellett: Point taken - let's leave this section as is :+1:

DonaldKellett (approved):

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+- The author might think that header files for the testing framework are included automatically by the code runner. That is not the case though, and test suites need to include `criterion/criterion.h` explicitly.
+
+### Compilation warnings
+
+Compiler options related to warnings used by the C runner are somewhat strict and require some discipline to get the code to compile cleanly. `-Wall` and `-Wextra` may cause numerous warnings and some of them are very pedantic. However, code of C kata should still compile cleanly, without any warnings logged to the console. Even when a warning does not cause any problem with tests, users get distracted by them and blame them for failing tests.
```
kazk, line 68: Maybe mention that you can disable warnings with pragmas if it doesn't make sense to fix.

kazk (approved):
Looks great! Thanks for writing this.

content/languages/c/authoring/index.md
hobovsky: Oh, haven't thought of that. I wonder if we want to mention it, though. Quality of C code in kata is, IMO, somewhat low, and people just silencing the warnings won't help improving it, I am afraid :( But maybe it's just me being too pessimistic?

kazk: I'd prefer authors fixing them too because that usually makes updating easier. So I'm fine with not mentioning it.

ggorlen (approved):

content/languages/c/authoring/index.md
error256, line 77: `'You won!'` isn't just an enum-like `'won'` (or should it be `'win'` for consistency with `'draw'`?..), it's a formatted string to be shown, so it kind of is consistent with returning a `const char*`, but creating a user-friendly string isn't really something that the function should do. So the type probably should be changed, but it isn't because C is different somehow.

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+### Reference solution
+
+If the test suite happens to use a reference solution to calculate expected values (which [should be avoided](/authoring/guidelines/submission-tests/#reference-solution) when possible), or some kind of reference data like precalculated arrays, etc., it must not be possible for the user to call it, redefine, overwrite or directly access its contents. To prevent this, it should be defined as `static` in the tests implementation file.
+
+The reference solution or data ___must not___ be defined in the [Preloaded code](/authoring/guidelines/preloaded/).
+
+
+### Redeclaration of user solution
+
+The solution function should be re-declared in the file with submission tests. Such re-declaration prevents a compilation warning about implicitly declared functions, and additionally stops users from tampering with the prototype of the solution function, for example, to remove const-ness of parameters, or change types of parameters, etc.
```
error256, line 116: If it's not yet declared in the file, it isn't really a *re-*declaration...
Implicit declarations assume everything is `int`, so warnings isn't the biggest concern, although errors caused by this may often be not noticeable on x64.
It doesn't prevent changing the prototype by the user, it's a totally different compilation unit. Like those solutions that define `const char[]` instead of a function, the compiler doesn't care.

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+#include <criterion/criterion.h>
+
+//include all required headers
+#include <math.h>
+#include <time.h>
+#include <stdlib.h>
+#include <stddef.h>
+
+//redeclare the user solution
+void square_every_item(double items[], int size);
```
error256, line 163: The type of the size should be `size_t`.

content/languages/c/authoring/index.md
```diff
@@ -0,0 +1,285 @@
...
+                         input[invalid_position], 
+                         reference[invalid_position], 
+                         actual[invalid_position]);
+  }
+}
+
+//a set of large random tests, with not so detailed debugging messages
+Test(random_tests, large_arrays) {
+  
+  double array[1000];
```
error256, line 265: Stack-allocated. OK as long as it works, but it won't work if the arrays are too big.

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -0,0 +1,24 @@
...
+
+# Memory Management in C kata
+
+_TBD_
+
+## Arrays and strings
+
+- malloc in solution and free in tests
+- pass in a preallocated buffer (use size hints if possible)
+- two functions: get size, allocate in tests, run solution
```
error256, line 15: Is this approach used anywhere? The function can return the size when it gets NULL as the buffer, so how is two function better?

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -0,0 +1,24 @@
...
+- two functions: solution with allocation, deallocation. Bookkeeping information managed by user or passed as additional `void*`
+- one function: accept buffer+size, return retsult or error and required size
+- avoid string constants, use named symbols
+- reporting size:output param, structure, sentinel teminators
+## Two-dimensional arrays
+
+- N+1 allocations
+- Flat
+- TOC + Flat
```
error256, line 24: TOC? I don't know or don't remember what it means...

error256 (changes requested):

content/languages/c/authoring/index.md
hobovsky: Please see [this comment](https://github.com/codewars/docs/pull/228#discussion_r554881117) and [this comment](https://github.com/codewars/docs/pull/228#discussion_r554930522). Maybe my example was not really fortunate, because I specifically meant kata with interface like, for example, "Red Knight", which says "return the color of captured pawn" while only "White" and "Black" are possible, or "Connect Four" with return values of "Plyer_Red" and "Player_Green" etc. I will think how to improve the example.

content/languages/c/authoring/index.md
hobovsky: It's true that it's different compilation unit, but when the user solution is (re)declared in tests and user tampers with its prototype in the solution file, the linker seems to complain, preventing it from running.

EDIT: okay I see it does not work as I thought it does, and I guess I will remove this part.

content/languages/c/authoring/index.md
hobovsky: 👍

content/languages/c/authoring/memory-management-techniques.md
hobovsky: This part is worked on in PR #231 which is still in draft, but you can review what's available there, would really appreciate.


> Is this approach used anywhere?

Yes, I think I've encountered it in some libraries, but my C got really rusty so I might remember some things not as they really were :)

> The function can return the size when it gets NULL as the buffer, so how is two function better?

And what if interface of the solution function has no good place to return the size? Should it use `size_t*` ans an in-out param and fill it when invoked with NULL, and read when called with non-null?

content/languages/c/authoring/memory-management-techniques.md
hobovsky: "TOC" as in "table of contents", I made the name up and I do not know how (if) it is professionally called: two malloced buffers, one flat array of `T*` and another one of `T**` with elements pointing to "rows" in the first one.

This part is not ready yet, but if you'd like to discuss it, please use dedicated thread #230 or PR #231

content/languages/c/authoring/index.md
hobovsky: I tried to change the example so it would stress the point better. Do you think it's better now?

error256 (approved):

Issue #230, opened 2021-01-10
Memory management patterns for C kata


hobovsky:
Create a descriptions of patterns for memory management which would familiarize C authors with any approach other than "malloc everything".

hobovsky:
I will move the discussion here, because I think it's difficult to discuss things in PR:

@ggorlen commented (https://github.com/codewars/docs/pull/231#pullrequestreview-566731480):

> Interesting article with a lot of potential! While it offers some good food for thought and a few novel ideas I hadn't considered, my take is that it's a bit too opinionated and complicated for its own good.
> 
> In practice, in the majority of the C kata I've seen that require dynamic memory allocation, the solution function allocates its own memory with `malloc` and returns a pointer for the caller to free (the "naive" strategy discussed here). For LeetCode, where nearly all of the problems are translated into C, this strategy is applied across the board. The format consistency makes it easy to focus on the solution right away, and even with 2- or 3-d arrays, I've never found it unduly difficult/burdensome to `malloc` and return inner arrays and an array of sizes in the case of jagged arrays (as in this [example](https://leetcode.com/problems/n-ary-tree-level-order-traversal/)).
> 
> That said, I agree with the guide that this "naive" approach isn't always better, both for kata and software engineering in general (important to consider if only to the extent that we don't want to recommend unusual/esoteric/questionable stuff), but it seems to be fine for most kata in practice. The preallocated buffer strategy is fine too and is a great fit for many kata types, but I'm not sure it's helpful to pick sides. Making authors aware of the characteristics of each without offering an opinion seems ideal.
> 
> As for the flattened nd arrays, reallocation and other approaches, these seem overly-complex. I can't imagine too many kata will need these. It's worth mentioning the flattened array approach because it exists, may be suitable for the occasional challenge and seems to align well with assembly language translations (I'm not sure, I've never done asm on CW), but is unlikely to mesh very well with typical translations coming from high-level languages that are overwhelmingly going to use multi-level nd arrays.

hobovsky:
### TL;DR

1. I have a specific problem of clueless authors creating bad content which I am trying to solve. I would appreciate any suggestions how to approach it.
1. I agree that things are complex, but I believe the complexity is an inherent part and an essential feature of the problem and I do not know if/how it could be avoided. I would be glad to be proven wrong, or explained how to present things in an easy way.

[...]

PR #231, opened 2021-01-10
C memory management

hobovsky:
Dedicated issue: #230 
Preview: [here](https://deploy-preview-231--reverent-edison-2864ea.netlify.app/languages/c/authoring/memory-management-techniques/)

As I explained to Unnamed below:

> [...] my intention [behind this article] is not to create C programmers guide, I am just looking for a way to transfer some ideas to C authors and translators, to get rid of ideas like requirement of mallocing a 3-element array to return hours, minutes, and seconds. Or returning `Position**`, where each entry is... a pointer to an individually malloced, single `Position` :( And there are even worse. I just thought about collecting some ideas and throw them at the authors, so they could pick something suitable for their kata. It's meant to be CW- and kata-oriented, and not an industry grade HOWTO ;) That's why I do not want to make things too complex and too detailed. Maybe I already have, maybe this paragraph, or this whole article, is simply not necessary?

Let me know what you think about this article, and even if you think it's not necessary or too detailed, it was a real fun writing it ;)


content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,341 @@ sidebar: "language:c"
...
+        //release the memory after the test
+        free(array);
+    }
+}
+```
+
+This approach is used when the size of the answer cannot be easily inferred by the test suite, but can be efficiently calculated by the user, potentially without the overhead of calculating the actual solution.
+
+
+#### When the size is not known, and difficult to calculate: assume (or guess) the initial size and reallocate if too small
```
error256, line 203: This seem like a valid approach, but the fact that the size isn't easily predictable probably means that the task is relatively high-level and the solution is heavy enough that reallocating memory there isn't that bad, while recalculating the result after the buffer was too small is...

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,341 @@ sidebar: "language:c"
...
+
+
+
+### Flat array
+
+Very often overlooked, but a very good approach to represent 2D arrays is to store them in a regular, linear array of `T[ ]`. It can be effectively used when bounds between inner arrays can be efficiently determined, for example each row of a matrix has a well known length, rows of a Pascal's triangle have precisely defined, although different, lengths, and string entries are clearly terminated.
+
+This way, complexity of memory management is greatly reduced, because whole necessary memory can be allocated and freed with a single call to `malloc` (or equivalent) and `free`.
+
+Drawbacks of this approach is that the interface of the solution does not resemble its logical structure, i.e. elements of a matrix cannot be accessed with, for example, `matrix[row][col]`, but with `matrix[row * size + col]`.
```
error256, line 339: `matrix[row][col]` can be used, it just require a type cast. C99 is more than 20 years old, so it's not even _modern_ C, it's just C.

content/languages/c/authoring/memory-management-techniques.md
hobovsky: This approach is definitely used, but in various variants. Works well for functions which just fetch data, but I've seen it also in APIs for digital signature and encryption. Callee can keep track of partial progress on their own and reuse it, or they can allocate memory locally, drive the process to its end, and upon return signal the required amount of memory by out param and copy potentially truncated amount of data to the output buffer, etc. Possibilities are many.

It's used by quite a number of Win32 API functions, for example for [reading from the Windows registry](https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-reggetvaluea) (in particular, see the note on `HKEY_PERFORMANCE_DATA`), but as it's said, characteristic of that function is _not_ to perform heavy computations.

Nonetheless, my intention is not to create C programmers guide, I am just looking for a way to transfer some ideas to C authors and translators, to get rid of ideas like requirement of mallocing a 3-element array to return hours, minutes, and seconds. Or returning `Position**`, where each entry is... a pointer to an individually malloced, single `Position` :( And there are even worse. I just thought about collecting some ideas and throw them at the authors, so they could pick something suitable for their kata. It's meant to be CW- and kata-oriented, and not an industry grade HOWTO ;) That's why I do not want to make things too complex and too detailed. Maybe I already have, maybe this paragraph, or this whole article, is simply not necessary?

error256: > if the whole point is to delegate allocations to the caller and it does not even know how the buffer was allocated

I feel like it's not too beneficial in this case. The reasons to do this may be:
* avoid reallocations if a function is called many times... it's not like it's called _millions_ of times on CW;
* allow using different allocation methods, especially in cross-platform code... is it worth it here?
So it's not like it's bad, but I'm not sure whether it should be in the tutorial. Those who are ready to read looong and complicated (relative to their level) tutorials may as well google themselves, while others would probably already be confused by this point. Anyway, I'm not the target audience, so I don't know what they need.

content/languages/c/authoring/memory-management-techniques.md
hobovsky: Oh, this is interesting. Could you show me then how it would work in such case then?

```c
int w = rand() % 100 + 10, h = rand() % 50 + 10;
char* maze_buffer = malloc(w * h + 1);
generate_maze(maze_buffer, w, h); //invoke user solution

//how can I cast the linear maze_buffer to 2D maze?
??? maze = (???)maze_buffer;  //how to cast it?
char cell = maze[5][3];

//....
free(maze_buffer);
```

I stopped working with C before 2005, so even C99 was not widespread across compilers back then :D Thanks!

content/languages/c/authoring/memory-management-techniques.md
hobovsky: Ok, I think I get what you mean, thanks. I will think what to do with this paragraph, but I think in the end I will just remove it.

Do you think other parts are more or less OK? Should be simplified? Or maybe the whole article is pointless as no one would benefit from it?

error256: I haven't read it all thoroughly, I'm not sure.

content/languages/c/authoring/memory-management-techniques.md
error256: `char (*maze)[w] = (char(*)[w])maze_buffer;` for an existing buffer, just `char maze[h][w];` for allocation on stack, `void generate_maze(size_t w, size_t h, char maze_buffer[h][w])` as a function argument (dimensions have to be before the array), asterisks may be used in prototypes like `void generate_maze(size_t, size_t, char maze_buffer[*][*]` when there are no named arguments.

hobovsky: Oh great, thank you. I was not aware that it's possible to use variables in type specifiers. Smarter every day!

hobovsky: I cannot get to declaring the return type of the function. I think it has to stay as `char*` and casts have to be done by the party accessing it, i.e. by solution to fill in, and by the test suite to read it?

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+
+This approach is used when the size of the answer cannot be easily inferred by the test suite, but can be efficiently calculated by the user, potentially without the overhead of calculating the actual solution.
+
+
+#### When the size is not known, and difficult to calculate: assume (or guess) the initial size and reallocate if too small
+
+This approach is a combination of the two above. It has a somewhat complex interface, but allows for a performance compromise when the size of the result is not known upfront, and cannot be efficiently estimated without performing actual calculations. General scheme is that the test suite passes in some preallocated buffer, and when solution determines that the buffer is too small, it reports an error. The tests can use some strategy to grow the buffer and retry the solution. When the call to solution succeeds, it fills the buffer with the result and reports its size.
+
+:::info
+This paragraph is probably too complex and not suitable for Codewars kata. It will be probably removed.
```
ggorlen, line 218: +1 on removing this.

ggorlen (commented):
Interesting article with a lot of potential! While it offers some good food for thought and a few novel ideas I hadn't considered, my take is that it's a bit too opinionated and complicated for its own good.

In practice, in the majority of the C kata I've seen that require dynamic memory allocation, the solution function allocates its own memory with `malloc` and returns a pointer for the caller to free (the "naive" strategy discussed here). For LeetCode, where nearly all of the problems are translated into C, this strategy is applied across the board. The format consistency makes it easy to focus on the solution right away, and even with 2- or 3-d arrays, I've never found it unduly difficult/burdensome to `malloc` and return inner arrays and an array of sizes in the case of jagged arrays (as in this [example](https://leetcode.com/problems/n-ary-tree-level-order-traversal/)).

That said, I agree with the guide that this "naive" approach isn't always better, both for kata and software engineering in general (important to consider if only to the extent that we don't want to recommend unusual/esoteric/questionable stuff), but it seems to be fine for most kata in practice. The preallocated buffer strategy is fine too and is a great fit for many kata types, but I'm not sure it's helpful to pick sides. Making authors aware of the characteristics of each without offering an opinion seems ideal.

As for the flattened nd arrays, reallocation and other approaches, these seem overly-complex. I can't imagine too many kata will need these. It's worth mentioning the flattened array approach because it exists, may be suitable for the occasional challenge and seems to align well with assembly language translations (I'm not sure, I've never done asm on CW), but is unlikely to mesh very well with typical translations coming from high-level languages that are overwhelmingly going to use multi-level nd arrays.

hobovsky:
@ggorlen, everyone, as I find it difficult to discuss things in a PR review, please let's continue the discussion to the issues board: https://github.com/codewars/docs/issues/230#issuecomment-759085853

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+
+    //after performing all necessary assertions,
+    //free the array allocated by the user solution
+    free(actual);
+}
+```
+
+</details>
+
+This approach mimics the behavior of higher level languages, where functions are able to allocate and return arrays without problems. It seems a natural way for many authors, but, sometimes surplrisingly for them, it's often a bad one. It's often bad from the design point of view, but, even worse, in production setups it can be straight invalid and can lead to crashes.
```
DonaldKellett, line 71:
```suggestion
This approach mimics the behavior of higher level languages, where functions are able to allocate and return arrays without problems. It seems a natural way for many authors, but, sometimes surprisingly for them, it's often a bad one. It's often bad from a design point of view, but, even worse, in production setups it can be straight invalid and can lead to crashes.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+    //release the memory after all tests
+    free(array);
+}
+```
+
+</details>
+
+#### When the size is not known, but is easy to calculate: ask and allocate
+
+Another possible approach is to ask user to implement _two_ functions: one which would return the size needed to fit the result, and another one to perform actual operation. For example:
```
DonaldKellett, line 135:
```suggestion
Another possible approach is to ask the user to implement _two_ functions: one which would return the size needed to fit the result, and another one to perform the actual operation. For example:
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+
+<details>
+
+Kata task:
+
+> Given a positive integer `n`, return all terms of `n` top rows of Pascal's triangle, row by row, flattened into a single array.
```
DonaldKellett, line 141: Unless this is a direct quote from another source:

```suggestion
> Given a positive integer `n`, return all terms in the top `n` rows of Pascal's triangle flattened into a single array.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+```
+
+</details>
+
+This approach is used when the size of the answer cannot be easily inferred by the test suite, but can be efficiently calculated by the user, potentially without the overhead of calculating the actual solution.
+
+
+#### When the size is not known, and difficult to calculate: assume (or guess) the initial size and reallocate if too small
+
+This approach is a combination of the two above. It has a somewhat complex interface, but allows for a performance compromise when the size of the result is not known upfront, and cannot be efficiently estimated without performing actual calculations. General scheme is that the test suite passes in some preallocated buffer, and when solution determines that the buffer is too small, it reports an error. The tests can use some strategy to grow the buffer and retry the solution. When the call to solution succeeds, it fills the buffer with the result and reports its size.
```
DonaldKellett, line 215:
```suggestion
This approach is a combination of the two above. It has a somewhat complex interface, but allows for a performance compromise when the size of the result is not known upfront, and cannot be efficiently estimated without performing actual calculations. The general scheme is that the test suite passes in some pre-allocated buffer, and when the solution determines that the buffer is too small, it reports an error. The tests can then employ various strategies to grow the buffer and retry the solution. When the call to the solution succeeds, it fills the buffer with the result and reports its size.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+:::info
+This paragraph is probably too complex and not suitable for Codewars kata. It will be probably removed.
+:::
+
+<details>
+
+
+Kata task:
+
+> Given an interer `n > 1`, calculate Fibonacci numbers up to `n`.
```
DonaldKellett, line 226: Unless this is a direct quote from another source:

```suggestion
> Given an integer `n > 1`, calculate Fibonacci numbers up to `n`.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+    //release the memory after the test
+    free(array);
+}
+```
+
+</details>
+
+### Memory managed by the solution
+
+Opposite of memory managed by the test suite is the approach of pushing the responsibility to the user. This way, tests do not need to worry about problematic aspects of memory management, kata authors give freedom of implementation to users, and can reduce the boilerplate required to implement memory management.
```
DonaldKellett, line 333:
```suggestion
The opposite of managing memory in the test suite is the approach of delegating the responsibility to the solver. This way, tests do not need to worry about problematic aspects of memory management, kata authors give freedom of implementation to users, and can reduce the boilerplate required to implement memory management.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+
+
+#### Symmetric functions for allocation and deallocation
+
+This idea basically boils down to asking users to provide their equivalents of allocation and deallocation functions. Solution function is responsible not only for solving the task, but also for allocation of the memory, and storing of book-keeping information. Clean-up function is responsible for releasing resources.
```
DonaldKellett, line 338:
```suggestion
This idea basically boils down to asking users to provide their equivalents of allocation and de-allocation functions. The solution function is responsible not only for solving the task, but also for allocation of memory and storing of book-keeping information. The clean-up function is responsible for releasing resources.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+
+There's many possible ways of implementing the allocation scheme and corresponding clean-up function, and its usage usually ends up being similar to the [naive approach](#naive-approach-malloc-in-the-solution-and-free-in-tests) described in the beginning. Example implementation could be similar to:
```
DonaldKellett, line 340:
```suggestion
There are many possible ways of implementing the allocation scheme and corresponding clean-up function, and its usage usually ends up being similar to the [naive approach](#naive-approach-malloc-in-the-solution-and-free-in-tests) described in the beginning. An example implementation could be similar to:
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+
+<details>
+
+Kata task:
+
+> Given initial generation of a Game of Life population, return the state and the size of the game world after `n` generations.
```
DonaldKellett, line 346: Unless this is a direct quote from another source:

```suggestion
> Given the initial generation of a Game of Life population, return the state and size of the game world after `n` generations.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+
+As this approach very useful for more complex memory structures, a couple of examples can be found in the section on [Two-dimensional arrays](#two-dimensional-arrays).
+
+
 ## Two-dimensional arrays
 
-- N+1 allocations
-- Flat
-- TOC + Flat
+Some kata require the user solution to return a two-dimensional array, for example a 2D matrix, or an array of C-strings. Such scenarios are a bit more complex, because not only the higher order array has to be properly managed, but all its individual entries as well. Exact approach selected for allocation of such structure depends on the scenario, because different techniques are suitable for square or rectangular arrays, jagged arrays, arrays of null-terminated strings, etc.
```
DonaldKellett, line 394:
```suggestion
Some kata require the user solution to return a two-dimensional array, for example a 2D matrix, or an array of C-strings. Such scenarios are a bit more complex, because not only does the higher order array have to be properly managed, but all its individual entries as well. The exact approach selected for allocation of such structures depend on the scenario, because different techniques are suitable for square or rectangular arrays, jagged arrays, arrays of null-terminated strings, etc.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+
+
+Memory for 2D arrays can be managed both by test suite, or the user solution. As long as the size of the 2D array is known before calling a solution and does not change through the course of calculations, test suite can choose to perform all necessary allocations, and pass the memory to the solution function ready to use. This is a very good approach when working with chessboards, sudokus, matrices and mazes of predetermined sizes, etc. However, when it would be the case that the size of the answer cannot be easily determined beforehand, the technique with clean-up function provided by the user turns out to be helpful. User provided clean-up function is used in the examples below, but authors can choose to manage the memory in the test suite if it fits the task of their kata.
```
DonaldKellett, line 397:
```suggestion
Memory for 2D arrays can be managed by the test suite or the user solution. As long as the size of the 2D array is known before calling a solution and does not change through the course of calculations, the test suite can choose to perform all necessary allocations and pass the memory to the solution function ready to use. This is a very good approach when working with chessboards, sudokus, matrices and mazes of predetermined sizes, etc. However, in the case that the size of the answer cannot be easily determined beforehand, the technique with clean-up function provided by the user turns out to be helpful. A user provided clean-up function is used in the examples below, but authors can choose to manage the memory in the test suite if it fits the task of their kata.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+
+
+### Naive approach: N+1 allocations
+
+This is the most common, and also the worst possible approach of using dynamically allocated memory. It tends to be slow, causes excessive memory fragmentation, and is usually inferior to available alternatives.
```
DonaldKellett, line 402:
```suggestion
This is the most common yet worst possible approach of using dynamically allocated memory. It tends to be slow, causes excessive memory fragmentation and is usually inferior to available alternatives.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+}
+```
+
+</details>
+
+### Flat array
+
+Very often overlooked, but a very good approach to represent 2D arrays is to store them in a regular, linear array of `T[ ]`. It can be effectively used when bounds between inner arrays can be efficiently determined, for example each row of a matrix has a well known length, rows of a Pascal's triangle have precisely defined, although different, lengths, and string entries are clearly terminated.
+
+This way, complexity of memory management is greatly reduced, because whole necessary memory can be allocated and freed with a single call to `malloc` (or equivalent) and `free`.
```
DonaldKellett, line 437:
```suggestion
This way, the complexity of memory management is greatly reduced since all necessary memory can be allocated and freed with a single call to `malloc` (or equivalent) and `free`.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+
+Drawbacks of this approach is that the interface of the solution does not resemble its logical structure, i.e. elements of a matrix cannot be accessed with, for example, `matrix[row][col]`, but with `matrix[row * size + col]`. It also fits best rectangular arrays (i.e. arrays with equal length of all rows).
```
DonaldKellett, line 439:
```suggestion
Drawbacks of this approach include:

- The solution does not resemble its logical structure, e.g. elements of a matrix cannot be accessed with, for example, `matrix[row][col]`, but with `matrix[row * size + col]`
- It is best suited for perfectly rectangular arrays, i.e. arrays whose sub-arrays all have equal length
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+    free(world[0]);
+
+    //deallocate the smaller buffer
+    free(world_rows);
+}
+```
+
+</details>
+
+This method is somewhat problematic when the width of the internal arrrays is a subject to change thorough the calculations. While both buffers can be easily reallocated to grow or shrink (for example to add new rows), changing the width of the array causes that the data in rows needs to be manually "shifted apart", and entries in the rows array need to be updated.
```
DonaldKellett, line 509:
```suggestion
This method is somewhat problematic when the width of the internal arrays is subject to change through calculations. While both buffers can be easily reallocated to grow or shrink (for example to add new rows), changing the width of the array requires the data in rows to be manually "shifted apart", and entries in the affected rows need to be updated.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,506 @@ sidebar: "language:c"
...
+
+This approach is also requires additional book-keeping when used for jagged arrays, unless entries of adjacent rows are clearly separated (as it happens for, for example, an array of C-strings).
```
DonaldKellett, line 511:
```suggestion
This approach also requires additional book-keeping when used for jagged arrays, unless entries of adjacent rows are clearly separated, e.g. as it happens for an array of C-strings.
```

DonaldKellett (changes requested):

content/languages/c/authoring/memory-management-techniques.md
error256: Actually, both dimensions can be in the type, just as with regular arrays: `char (*maze)[h][w] = (char(*)[h][w])maze_buffer;`. As for returning it, I don't think it's possible, it still is a low level language.

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,509 @@ sidebar: "language:c"
...
 ## Arrays and strings
 
-- malloc in solution and free in tests
-- pass in a preallocated buffer (use size hints if possible)
-- two functions: get size, allocate in tests, run solution
-- two functions: solution with allocation, deallocation. Bookkeeping information managed by user or passed as additional `void*`
-- one function: accept buffer+size, return retsult or error and required size
-- avoid string constants, use named symbols
-- reporting size:output param, structure, sentinel teminators
+Since C-strings and arrays of other types are similar from the perspective of memory management, this article uses examples of integer arrays. However most of the techniques presented here applies equally to handling memory holding integers, floats, characters, zero-terminated or not.
```
Steffan153, line 23:
```suggestion
Since C-strings and arrays of other types are similar from the perspective of memory management, this article uses examples of integer arrays. However, most of the techniques presented here apply equally to handling memory holding integers, floats, and characters, zero-terminated or not.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,509 @@ sidebar: "language:c"
...
+
+    //after performing all necessary assertions,
+    //free the array allocated by the user solution
+    free(actual);
+}
+```
+
+</details>
+
+This approach mimics the behavior of higher level languages, where functions are able to allocate and return arrays without problems. It seems a natural way for many authors, but, sometimes surprisingly for them, it's often a bad one. It's often bad from a design point of view, but, even worse, in production setups it can be straight invalid and can lead to crashes.
```
Steffan153, line 71:
```suggestion
This approach mimics the behavior of higher-level languages, where functions can allocate and return arrays without problems. It seems a natural way for many authors, but, sometimes surprisingly for them, it's often a bad one. It's often bad from a design point of view, but, even worse, in production setups, it can be straight invalid and can lead to crashes.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,509 @@ sidebar: "language:c"
...
+
+The biggest problem with this philosophy is that the test suite does not always know how much memory the solution would need to fit all the requested results in. But there are a few possible ways to resolve this issue.
+
+
+#### When the size is known upfront: use preallocated buffer
+
+Sometimes it's perfectly known how large the result will be before the solution is called. For example, if the test suite asks to generate `n` Fibonacci numbers, it means that the resulting array needs to have the size of at least `n`. Sometimes the exact size is not known exactly, but it's possible to accurately estimate its upper bound. For example, a function which removes punctuation from a string needs to work on a buffer at least as large as an input string, but the result can turn out to be a bit smaller. In such cases, the test suite can allocate the buffer which would be big enough to keep the result, and pass it to the solution function:
```
Steffan153, line 83:
```suggestion
Sometimes it's perfectly known how large the result will be before the solution is called. For example, if the test suite asks to generate `n` Fibonacci numbers, it means that the resulting array needs to have the size of at least `n`. Sometimes the exact size is not known exactly, but it's possible to accurately estimate its upper bound. For example, a function that removes punctuation from a string needs to work on a buffer at least as large as an input string, but the result can turn out to be a bit smaller. In such cases, the test suite can allocate the buffer which would be big enough to keep the result, and pass it to the solution function:
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,509 @@ sidebar: "language:c"
...
+
+As this approach very useful for more complex memory structures, a couple of examples can be found in the section on [Two-dimensional arrays](#two-dimensional-arrays).
+
+
 ## Two-dimensional arrays
 
-- N+1 allocations
-- Flat
-- TOC + Flat
+Some kata require the user solution to return a two-dimensional array, for example a 2D matrix, or an array of C-strings. Such scenarios are a bit more complex, because not only does the higher order array have to be properly managed, but all its individual entries as well. The exact approach selected for allocation of such structures depend on the scenario, because different techniques are suitable for square or rectangular arrays, jagged arrays, arrays of null-terminated strings, etc.
```
Steffan153, line 394:
```suggestion
Some kata require the user solution to return a two-dimensional array, for example, a 2D matrix, or an array of C-strings. Such scenarios are a bit more complex, because not only does the higher-order array have to be properly managed, but all its individual entries as well. The exact approach selected for the allocation of such structures depends on the scenario because different techniques are suitable for square or rectangular arrays, jagged arrays, arrays of null-terminated strings, etc.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,509 @@ sidebar: "language:c"
...
+
+
+Memory for 2D arrays can be managed by the test suite or the user solution. As long as the size of the 2D array is known before calling a solution and does not change through the course of calculations, the test suite can choose to perform all necessary allocations and pass the memory to the solution function ready to use. This is a very good approach when working with chessboards, sudokus, matrices and mazes of predetermined sizes, etc. However, in the case that the size of the answer cannot be easily determined beforehand, the technique with clean-up function provided by the user turns out to be helpful. A user provided clean-up function is used in the examples below, but authors can choose to manage the memory in the test suite if it fits the task of their kata.
```
Steffan153, line 397:
```suggestion
Memory for 2D arrays can be managed by the test suite or the user solution. As long as the size of the 2D array is known before calling a solution and does not change through the course of calculations, the test suite can choose to perform all necessary allocations and pass the memory to the solution function ready to use. This is a very good approach when working with chessboards, sudokus, matrices and mazes of predetermined sizes, etc. However, in the case that the size of the answer cannot be easily determined beforehand, the technique with a clean-up function provided by the user turns out to be helpful. A user-provided clean-up function is used in the examples below, but authors can choose to manage the memory in the test suite if it fits the task of their kata.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,509 @@ sidebar: "language:c"
...
+        free(world[i]);
+    free(world);
+}
+```
+
+</details>
+
+### Flat array
+
+Very often overlooked, but a very good approach to represent 2D arrays is to store them in a regular, linear array of `T[ ]`. It can be effectively used when bounds between inner arrays can be efficiently determined, for example each row of a matrix has a well known length, rows of a Pascal's triangle have precisely defined, although different, lengths, and string entries are clearly terminated.
```
Steffan153, line 435:
```suggestion
Very often overlooked, but a very good approach to represent 2D arrays is to store them in a regular, linear array of `T[ ]`. It can be effectively used when bounds between inner arrays can be efficiently determined, for example, each row of a matrix has a well-known length, rows of a Pascal's triangle have precisely defined, although different, lengths, and string entries are clearly terminated.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,509 @@ sidebar: "language:c"
...
+</details>
+
+### Memory managed by the solution
+
+The opposite of managing memory in the test suite is the approach of delegating the responsibility to the solver. This way, tests do not need to worry about problematic aspects of memory management, kata authors give freedom of implementation to users, and can reduce the boilerplate required to implement memory management.
+
+
+#### Symmetric functions for allocation and deallocation
+
+This idea basically boils down to asking users to provide their equivalents of allocation and de-allocation functions. The solution function is responsible not only for solving the task, but also for allocation of memory and storing of book-keeping information. The clean-up function is responsible for releasing resources.
```
Steffan153, line 338:
```suggestion
This idea boils down to asking users to provide their equivalents of allocation and de-allocation functions. The solution function is responsible not only for solving the task but also for allocation of memory and storing of book-keeping information. The clean-up function is responsible for releasing resources.
```

Steffan153 (changes requested):

Steffan153 (approved):

hobovsky:
I am moving this back to draft and will try to simplify it. Let's see what I can come up with.

DonaldKellett (approved):

hobovsky:
I've reorganized the article, please let me know if it's better now.

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+### Arrays and strings
+
+Since C-strings and arrays of other types are similar from the perspective of memory management, this article uses examples of integer arrays. However, most of the techniques presented here apply equally to handling memory holding integers, floats, and characters, zero-terminated or not.
+
+
+## Memory Management Patterns
+
+In C, unlike for example Python, Java, C# or Javascript, dynamically allocated memory is not managed by the runtime. It's considered to be an external resource, just like a file, a DB or network connection, or a hardware device. The program itself has to take care of it properly, allocating it when necessary, and freeing when no longer needed.
+
+In a kata, the memory can be managed either by the test suite, by the user, or both. Authors can choose the way how their kata should deal with memory and ownership strategy their kata should use. However, they should be aware of the advantages and disadvantages of each such strategy, and when and which applies the best.
```
Steffan153:
```suggestion
In kata, the memory can be managed either by the test suite, by the user, or both. Authors can choose the way how their kata should deal with memory and the ownership strategy their kata should use. However, they should be aware of the advantages and disadvantages of each such strategy, and when and which applies the best. 
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+
+
+### Statically allocated constant data
+
+The best way to avoid problems with memory allocation is to avoid unnecessary memory allocation. This advice might sound tricky, but there are simply many kata which require dynamic memory allocation or operation on data pointed by pointers, while it's not necessary and could be avoided. One commonly occuring example of such situation is when a kata requires returning a pointer to a string which could be replaced by a constant, which is used in particular when translating kata from other languages. Returning a string in high-level languages is not a problem, but in C it always raises questions of who should allocate it and how it should be allocated. Consider replacing the string with some simpler data type (eventually aliased with a `typedef`), and/or provide some symbolic constants for available values. For example, if the requirement for the JavaScript version is: _"Return the string 'BLACK' if a black pawn will be captured first, 'WHITE' if a white one, and 'NONE' if all pawns are safe."_, C version should preferably provide and use the named constants `BLACK`, `WHITE` and `NONE`.
```
Steffan153:
```suggestion
The best way to avoid problems with memory allocation is to avoid unnecessary memory allocation. This advice might sound tricky, but there are simply many kata that require dynamic memory allocation or operation on data pointed by pointers, while it's not necessary and could be avoided. One commonly occurring example of such a situation is when a kata requires returning a pointer to a string which could be replaced by a constant, which is used in particular when translating kata from other languages. Returning a string in high-level languages is not a problem, but in C it always raises questions of who should allocate it and how it should be allocated. Consider replacing the string with some simpler data type (eventually aliased with a `typedef`), and/or provide some symbolic constants for available values. For example, if the requirement for the JavaScript version is: _"Return the string 'BLACK' if a black pawn will be captured first, 'WHITE' if a white one, and 'NONE' if all pawns are safe."_, C version should preferably provide and use the named constants `BLACK`, `WHITE` and `NONE`. 
```
(yes, this does have a few changes, GH is not highlighting them for some reason)

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+  Player winner = who_won("B");
+
+  //constants can be asserted on with cr_assert_eq
+  cr_assert_eq(winner, NONE, "Expected: [%s], but was: [%s]", NONE, winner);
+
+  //...no clean-up necessary
+}
+```
+
+It is recommended to replace constant strings with some even simpler type, preferrably an `enum`, but if authors really want to stick to strings for some reason, they can use them.
```
Steffan153, line 104:
```suggestion
It is recommended to replace constant strings with some even simpler type, preferably an `enum`, but if authors really want to stick to strings for some reason, they can use them.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+
+Sometimes it's perfectly known how large the result will be before the solution is called, or it's possible to pre-allocate a buffer which will be large enough for every call. For example, if the test suite asks to generate `n` Fibonacci numbers, it means that the resulting array needs to have the size of at least `n`. Sometimes the exact size is not known exactly, but it's possible to accurately estimate its upper bound. For example, a function that removes punctuation from a string needs to work on a buffer at least as large as an input string, but the result can turn out to be a bit smaller. In such cases, the test suite can allocate the buffer which would be big enough to keep the result, and pass it to the solution function:
```
Steffan153, line 113:
```suggestion
Sometimes it's perfectly known how large the result will be before the solution is called, or it's possible to pre-allocate a buffer that will be large enough for every call. For example, if the test suite asks to generate `n` Fibonacci numbers, it means that the resulting array needs to have the size of at least `n`. Sometimes the exact size is not known exactly, but it's possible to accurately estimate its upper bound. For example, a function that removes punctuation from a string needs to work on a buffer at least as large as an input string, but the result can turn out to be a bit smaller. In such cases, the test suite can allocate the buffer which would be big enough to keep the result, and pass it to the solution function:
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+    }
+
+    //release the memory after all tests
+    free(array);
+}
+```
+
+</details>
+
+This technique is often overlooked by kata authors, but it greatly simplifies the way how user solutions are built and how they communicate with the test suite. User's solution does not have to worry about allocations or error handling, and can focus on its task. Test suite can use any allocation technique it wants, like automatic allocation on the stack, or dynamic allocation on a heap. Buffer can be allocated once and reused accros many test calls.
```
Steffan153, line 173:
```suggestion
This technique is often overlooked by kata authors, but it greatly simplifies the way how user solutions are built and how they communicate with the test suite. The user's solution does not have to worry about allocations or error handling, and can focus on its task. The test suite can use any allocation technique it wants, like automatic allocation on the stack, or dynamic allocation on a heap. Buffers can be allocated once and reused across many test calls.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+
+The biggest problem with allocated memory is that its size has to be known or possible to estimate before calling the user's solution. It's very often the case, but sometimes such estimation is not possible or easy. There are ways to work around this problem and work with memory allocated by the caller even when its size is not known upfront, but they are out of scope of this article. In such cases, kata can use a memory allocated by the user.
```
Steffan153, line 175:
```suggestion
The biggest problem with allocated memory is that its size has to be known or possible to estimate before calling the user's solution. It's very often the case, but sometimes such estimation is not possible or easy. There are ways to work around this problem and work with memory allocated by the caller even when its size is not known upfront, but they are out of the scope of this article. In such cases, kata can use a memory allocated by the user.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+
+
+### Mixed approach: `malloc` in the solution and `free` in tests
+
+In a vast majority of cases when a kata requires the solution to allocate memory, authors choose the naive approach of allocating the memory in the solution, and releasing it with `free` in the test suite after performing all necessary assertions. This mimics the behavior known from high-level languages where returning an array or object from inside of the user's solution is perfectly valid, but it's not always the best, or even correct, way of working with unmanaged memory in C.
+
+This approach is useful when the size of the result is not known before the call. The solution is reponsible for finding the correct size and returning it along with the pointer to the buffer itself, and the test suite is responsible for freeing it after every call.
```
Steffan153, line 182:
```suggestion
This approach is useful when the size of the result is not known before the call. The solution is responsible for finding the correct size and returning it along with the pointer to the buffer itself, and the test suite is responsible for freeing it after every call.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+
+<details>
+
+Kata task:
+
+> Given a natural nuber `n`, return all prime numbers up to and including `n`.
```
Steffan153, line 188:
```suggestion
> Given a natural number `n`, return all prime numbers up to and including `n`.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+    //free the array allocated by the user solution
+    free(actual);
+}
+```
+
+</details>
+
+This approach works in a way similar to functions like `strdup` or `asprintf`, which allocate required memory and pass its ownership to the caller. It's a good fit for Codewars kata because it's simple, effective, and works well in Codewars code runner.
+
+Potential issue with the mixed approach is not related to Codewars, but to "real world" C coding and design. It might not work well for complex memory structures, or when a callee has to do advanced book-keeping and tracking of allocated memory. It also does not work well when passing data between modules (for example, between libraries, or from a library to main program).
```
Steffan153, line 232:
```suggestion
A potential issue with the mixed approach is not related to Codewars, but to "real world" C coding and design. It might not work well for complex memory structures, or when a callee has to do advanced book-keeping and tracking of allocated memory. It also does not work well when passing data between modules (for example, between libraries, or from a library to the main program).
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+
+
+Just as any memory, 2D arrays can be managed by the test suite, or the user solution, or both. As long as the size of the 2D array is known before calling a solution and does not change through the course of calculations, the test suite can choose to perform all necessary allocations and pass the memory to the solution function ready to use. This is a very good approach when working with chessboards, sudokus, matrices and mazes of predetermined sizes, etc. However, in the case that the size of the answer cannot be easily determined beforehand, the mixed approach or the memory management by a callee with a clean-up function provided by the user can be better.
+
+:::note Note on examples
+For simplicity, this section uses terms "2D array", "array of arrays" and "matrix" interchangeably and assume row-major order, i.e. data can be accessed with `array[row][col]`.
```
Steffan153, line 303:
```suggestion
For simplicity, this section uses the terms "2D array", "array of arrays", and "matrix" interchangeably and assumes row-major order, i.e. data can be accessed with `array[row][col]`.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+:::
+
+### Naive approach: N+1 allocations
+
+This is the most common approach of using dynamically allocated multi-dimensional arrays. An array of pointers to rows is allocated first, and each row is allocated individually afterwrds.
```
Steffan153, line 308:
```suggestion
This is the most common approach of using dynamically allocated multidimensional arrays. An array of pointers to rows is allocated first, and each row is allocated individually afterwards.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+
+free(world);
+}
+```
+
+</details>
+
+This approach, although it seems to be simple, is affected by issues mostly related to performance. It tends to be slow, because every dynamic allocation requires a lookup of memory to be performed. It can also cause excessive memory fragmentation.
+
+Advantage of individually allocated rows is that ot works good for jagged arrays.
```
Steffan153:
```suggestion
The advantage of individually allocated rows is that it works well for jagged arrays.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+
+
+### Flat array
+
+Very often overlooked, but a very good approach to represent 2D arrays is to store them in a regular, linear array of `T[ ]`, potentially supported by some type casts between linear buffer and two-dimensional matrix.
```
Steffan153, line 449:
```suggestion
Very often overlooked, but a very good approach to represent 2D arrays is to store them in a regular, linear array of `T[ ]`, potentially supported by some type casts between a linear buffer and two-dimensional matrix. 
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+
+//deallocate all memory at once
+free(world_linear);
+```
+
+</details>
+
+This way, the complexity of memory management is greatly reduced since all necessary memory can be allocated and freed with a single call to `malloc` (or equivalent) and `free`.
+
+Drawback of the version with casts between linear and 2D array is that it is best suited for perfectly rectangular arrays, i.e. arrays whose sub-arrays all have equal length. However, the version without casts can be effectively used when bounds between inner arrays can be efficiently determined, for example, each row of a matrix has a well-known length, rows of a Pascal's triangle have precisely defined, although different, lengths, and string entries are clearly terminated.
```
Steffan153, line 490:
```suggestion
The drawback of the version with casts between linear and 2D arrays is that it is best suited for perfectly rectangular arrays, i.e. arrays whose sub-arrays all have equal length. However, the version without casts can be effectively used when bounds between inner arrays can be efficiently determined, for example, each row of a matrix has a well-known length, rows of a Pascal's triangle have precisely defined, although different, lengths, and string entries are clearly terminated.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,382 @@ sidebar: "language:c"
...
+
+This method also does not also fit perfectly the scenario when such array should be *returned* from a function. The function still has to specify its return type as `T*`, and the caller has to either work with linear form of the array, or perform the cast on its own.
```
Steffan153, line 492:
```suggestion
This method also does not fit perfectly the scenario when such an array should be *returned* from a function. The function still has to specify its return type as `T*`, and the caller has to either work with the linear form of the array or perform the cast on its own.
```

Steffan153 (changes requested):

Steffan153:
:roll_eyes: a new commit while I'm reviewing...

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,491 @@ sidebar: "language:c"
...
+### Arrays and strings
+
+Since C-strings and arrays of other types are similar from the perspective of memory management, most of the techniques presented here apply equally to handling memory holding integers, floats, and characters, zero-terminated or not.
+
+
+## Memory Management Patterns
+
+In C, unlike for example Python, Java, C# or Javascript, dynamically allocated memory is not managed by the runtime. It's considered to be an external resource, just like a file, a DB or network connection, or a hardware device. The program itself has to take care of it properly, allocating it when necessary, and freeing when no longer needed.
+
+In a kata, the memory can be managed either by the test suite, by the user, or both. Authors can choose the way how their kata should deal with memory and they can pick any ownership strategy. However, they should be aware of the advantages and disadvantages of each such strategy, and when and which applies the best.
```
Steffan153, line 34:
```suggestion
In kata, the memory can be managed either by the test suite, by the user, or both. Authors can choose the way how their kata should deal with memory and they can pick any ownership strategy. However, they should be aware of the advantages and disadvantages of each such strategy, and when and which applies the best. 
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,491 @@ sidebar: "language:c"
...
+
+
+### Statically allocated constant data
+
+The best way to avoid problems with memory allocation is to avoid unnecessary memory allocation. This advice might sound tricky, but there are simply many kata which require dynamic memory allocation or operation on data pointed by pointers, while it's simply not necessary and could be avoided. One commonly occuring example of such situation is when a kata requires returning a pointer to a string which could be replaced by a constant. It seems to appear particularly often when translating kata from other languages. Returning a string in high-level languages is not a problem, but in C it always raises questions of who should allocate it and how it should be allocated. Consider replacing the string with some simpler data type (eventually aliased with a `typedef`), and/or provide some symbolic constants for available values. For example, if the requirement for the JavaScript version is: _"Return the string 'BLACK' if a black pawn will be captured first, 'WHITE' if a white one, and 'NONE' if all pawns are safe."_, C version should preferably provide and use the named constants `BLACK`, `WHITE` and `NONE`.
```
Steffan153, line 39:
```suggestion
The best way to avoid problems with memory allocation is to avoid unnecessary memory allocation. This advice might sound tricky, but there are simply many kata that require dynamic memory allocation or operation on data pointed by pointers, while it's simply not necessary and could be avoided. One commonly occurring example of such a situation is when a kata requires returning a pointer to a string which could be replaced by a constant. It seems to appear particularly often when translating kata from other languages. Returning a string in high-level languages is not a problem, but in C it always raises questions of who should allocate it and how it should be allocated. Consider replacing the string with some simpler data type (eventually aliased with a `typedef`), and/or provide some symbolic constants for available values. For example, if the requirement for the JavaScript version is: _"Return the string 'BLACK' if a black pawn will be captured first, 'WHITE' if a white one, and 'NONE' if all pawns are safe."_, C version should preferably provide and use the named constants `BLACK`, `WHITE` and `NONE`. 
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,491 @@ sidebar: "language:c"
...
+for(int i=0; i < world_h; ++i)
+    free(world[i]);
+
+free(world);
+}
+```
+
+</details>
+
+Advantage of individually allocated rows is that it works good for jagged arrays.
```
Steffan153, line 341:
```suggestion
The advantage of individually allocated rows is that it works well for jagged arrays.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,491 @@ sidebar: "language:c"
...
+
+This approach, although it seems to be simple, is affected by issues mostly related to performance. It tends to be slow, because every dynamic allocation requires a lookup of memory to be performed. It can also cause excessive memory fragmentation.
+
+Additionally, it is sometimes unnecessarily used to return an array of data (usualy strings) which could be turned into constants.
```
Steffan153, line 345:
```suggestion
Additionally, it is sometimes unnecessarily used to return an array of data (usually strings) that could be turned into constants.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,491 @@ sidebar: "language:c"
...
+
+
+### Array of `const` data
+
+This approach is related to [returning a statically allocated const data](#statically-allocated-constant-data), but extended to arrays. Some kata require the user to return an array of strings, which coud be turned into constants. In such case, the array itself can be allocated dynamically, but its entries do not have to be.
```
Steffan153, line 350:
```suggestion
This approach is related to [returning a statically allocated const data](#statically-allocated-constant-data) but extended to arrays. Some kata require the user to return an array of strings, which could be turned into constants. In such a case, the array itself can be allocated dynamically, but its entries do not have to be.
```

Steffan153 (changes requested):

content/languages/c/authoring/memory-management-techniques.md
hobovsky: This point has been removed.

content/languages/c/authoring/memory-management-techniques.md
hobovsky: I added a note on casting between linear and 2d array.

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,494 @@ sidebar: "language:c"
...
+
+    //after performing all necessary assertions,
+    //free the array allocated by the user solution
+    free(actual);
+}
+```
+
+</details>
+
+This approach works in a way similar to functions like `strdup` or `asprintf`, which allocate required memory and pass its ownership to the caller. It's a good fit for Codewars kata because it's simple, effective, and works well in Codewars code runner.
```
DonaldKellett, line 235:
```suggestion
This approach works in a way similar to functions like `strdup` or `asprintf`, which allocate required memory and pass its ownership to the caller. It's a good fit for Codewars kata because it's simple, effective, and works well in Codewars' code runner.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,494 @@ sidebar: "language:c"
...
+
+A potential issue with the mixed approach is not related to Codewars, but to "real world" C coding and design. It might not work well for complex memory structures, or when a callee has to do advanced book-keeping and tracking of allocated memory. It also does not work well when passing data between modules (for example, between libraries, or from a library to the main program).
```
DonaldKellett, line 237: Please don't use "coding" as a synonym to "programming" ...

```suggestion
A potential issue with the mixed approach is not related to Codewars, but to "real world" C programming and design. It might not work well for complex memory structures, or when a callee has to do advanced book-keeping and tracking of allocated memory. It also does not work well when passing data between modules (for example, between libraries, or from a library to the main program).
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,494 @@ sidebar: "language:c"
...
+
+
+### Memory managed by the solution
+
+The opposite of managing memory in the test suite is the approach of delegating the responsibility to the solver. This way, tests do not need to worry about problematic aspects of memory management, kata authors give freedom of implementation to users, and can reduce the boilerplate required to implement memory management.
+
+This idea boils down to asking users to provide their equivalents of allocation and de-allocation functions. The solution function is responsible not only for solving the task but also for allocation of memory and storing of book-keeping information. The clean-up function is responsible for releasing resources.
+
+There are many possible ways of implementing the allocation scheme and corresponding clean-up function, but example implementation could be similar to:
```
DonaldKellett, line 246:
```suggestion
There are many possible ways of implementing the allocation scheme and corresponding clean-up function, but an example implementation could be:
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,494 @@ sidebar: "language:c"
...
-- Flat
-- TOC + Flat
+Some kata require the user solution to return a two-dimensional array, for example, a 2D matrix, or an array of C-strings. Such scenarios are a bit more complex, because not only does the higher-order array have to be properly managed, but all its individual entries as well. The exact approach selected for the allocation of such structures depends on the scenario because different techniques are suitable for square or rectangular arrays, jagged arrays, arrays of null-terminated strings, etc.
+
+
+Just as any memory, 2D arrays can be managed by the test suite, or the user solution, or both. As long as the size of the 2D array is known before calling a solution and does not change through the course of calculations, the test suite can choose to perform all necessary allocations and pass the memory to the solution function ready to use. This is a very good approach when working with chessboards, sudokus, matrices and mazes of predetermined sizes, etc. However, in the case that the size of the answer cannot be easily determined beforehand, the mixed approach or the memory management by a callee with a clean-up function provided by the user can be better.
```
DonaldKellett, line 306:
```suggestion
Just as any form of memory, 2D arrays can be managed by the test suite, user solution, or both. As long as the size of the 2D array is known before calling a solution and does not change through the course of calculations, the test suite can choose to perform all necessary allocations and pass the memory to the solution function ready to use. This is a very good approach when working with chessboards, sudokus, matrices and mazes of predetermined sizes, etc. However, in the case that the size of the answer cannot be easily determined beforehand, the mixed approach or memory management by the callee with a clean-up function provided by the user can be better.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,494 @@ sidebar: "language:c"
...
+
+free(world);
+}
+```
+
+</details>
+
+The advantage of individually allocated rows is that it works well for jagged arrays.
+
+This approach, although it seems to be simple, is affected by issues mostly related to performance. It tends to be slow, because every dynamic allocation requires a lookup of memory to be performed. It can also cause excessive memory fragmentation.
```
DonaldKellett, line 346:
```suggestion
This approach, despite appearing to be simple, is affected by issues mostly related to performance. It tends to be slow since each dynamic allocation requires a memory lookup. It can also cause excessive memory fragmentation.
```

DonaldKellett (changes requested):

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,85 +6,129 @@ sidebar: "language:c"
...
-## Arrays and strings
+### Arrays and strings
+
+Since C-strings and arrays of other types are similar from the perspective of memory management, most of the techniques presented here apply equally to handling memory holding integers, floats, and characters, zero-terminated or not.
+
+
+## Memory Management Patterns
 
-Since C-strings and arrays of other types are similar from the perspective of memory management, this article uses examples of integer arrays. However most of the techniques presented here applies equally to handling memory holding integers, floats, characters, zero-terminated or not.
+In C, unlike for example Python, Java, C# or Javascript, dynamically allocated memory is not managed by the runtime. It's considered to be an external resource, just like a file, a DB or network connection, or a hardware device. The program itself has to take care of it properly, allocating it when necessary, and freeing when no longer needed.
```
error256, line 32: I'm not sure about calling memory an "_external_ resource". Doesn't just "resource" convey the meaning already?

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,85 +6,129 @@ sidebar: "language:c"
...
 
-    //after performing all necessary assertions,
-    //free the array allocated by the user solution
-    free(actual);
+Test(fixed_tests, no_one_won) {
+ 
+  Player winner = who_won("B");
+
+  //constants can be asserted on with cr_assert_eq
+  cr_assert_eq(winner, NONE, "Expected: [%s], but was: [%s]", NONE, winner);
```
error256, line 99: Non-string assertion... Although it's supposed to be the same pointer, I think if it's not an enum for whatever reason, it should be treated as a string. Someone will return their own string literal eventually just to get rid of extra declarations in the code.

content/languages/c/authoring/memory-management-techniques.md
hobovsky: Is this better?

```suggestion
In C, unlike for example Python, Java, C# or Javascript, dynamically allocated memory is not managed by the runtime. It's considered to be a resource like any other, for example a file, a DB or network connection, or a hardware device. The program itself has to take care of it properly, allocating it when necessary, and freeing when no longer needed.
```

content/languages/c/authoring/memory-management-techniques.md
hobovsky: > Someone will return their own string literal eventually just to get rid of extra declarations in the code.

Fair point.

content/languages/c/authoring/memory-management-techniques.md
error256: I think it's better. Is it normal that I see raw `span`s in Suggested change?..

content/languages/c/authoring/memory-management-techniques.md
hobovsky: I removed constant strings completely, replaced them with enums.

content/languages/c/authoring/memory-management-techniques.md
hobovsky: Removed.

error256 (approved):

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,378 @@ sidebar: "language:c"
...
+
+
+### Arrays and strings
+
+Since C-strings and arrays of other types are similar from the perspective of memory management, most of the techniques presented here apply equally to handling memory holding integers, floats, and characters, zero-terminated or not.
+
+
+## Memory Management Patterns
+
+In C, unlike for example Python, Java, C# or Javascript, dynamically allocated memory is not managed by the runtime. It's considered to be a resource like any other, for example a file, a DB or network connection, or a hardware device. The program itself has to take care of it properly, allocating it when necessary, and freeing when no longer needed.
```
Steffan153, line 32:
```suggestion
In C, unlike for example Python, Java, C#, or Javascript, dynamically allocated memory is not managed by the runtime. It's considered to be a resource like any other, for example, a file, a DB or network connection, or a hardware device. The program itself has to take care of it properly, allocating it when necessary, and freeing when no longer needed.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,378 @@ sidebar: "language:c"
...
+
+
+### Statically allocated constant data
+
+The best way to prevent problems with memory allocation is to avoid unnecessary memory allocation. This advice might sound tricky, but there are simply many kata that require dynamic memory allocation or operation on data pointed by pointers, while it's simply not necessary and could be avoided. One commonly occurring example of such a situation is when a kata requires returning a pointer to a string which could be replaced by a constant. It seems to appear particularly often when translating kata from other languages. Returning a string in high-level languages is not a problem, but in C it always raises questions of who should allocate it and how it should be allocated. Consider replacing the string with an `enum`. For example, if the requirement for the JavaScript version is: _"Return the string 'BLACK' if a black pawn will be captured first, 'WHITE' if a white one, and 'NONE' if all pawns are safe."_, the C version should preferably provide and use the named constants `BLACK`, `WHITE` and `NONE`.
```
Steffan153, line 39:
```suggestion
The best way to prevent problems with memory allocation is to avoid unnecessary memory allocation. This advice might sound tricky, but there are simply many kata that require dynamic memory allocation or operation on data pointed by pointers, while it's simply not necessary and could be avoided. One commonly occurring example of such a situation is when a kata requires returning a pointer to a string which could be replaced by a constant. It seems to appear particularly often when translating kata from other languages. Returning a string in high-level languages is not a problem, but in C it always raises questions of who should allocate it and how it should be allocated. Consider replacing the string with an `enum`. For example, if the requirement for the JavaScript version is: _"Return the string 'BLACK' if a black pawn will be captured first, 'WHITE' if a white one, and 'NONE' if all pawns are safe."_, the C version should preferably provide and use the named constants `BLACK`, `WHITE`, and `NONE`.
```

Steffan153 (approved):

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,378 @@ sidebar: "language:c"
...
+    //release the memory after all tests
+    free(array);
+}
+```
+
+</details>
+
+This technique is often overlooked by kata authors, but it greatly simplifies the way how user solutions are built and how they communicate with the test suite. The user's solution does not have to worry about allocations or error handling, and can focus on its task. The test suite can use any allocation technique it wants, like automatic allocation on the stack, or dynamic allocation on a heap. Buffers can be allocated once and reused across many test calls.
+
+The biggest problem with allocated memory is that its size has to be known or possible to estimate before calling the user's solution. It's very often the case, but sometimes such estimation is not possible or easy. There are ways to work around this problem and work with memory allocated by the caller even when its size is not known upfront, but they are out of the scope of this article. In such cases, kata can use a memory allocated by the user.
```
DonaldKellett, line 158:
```suggestion
The biggest problem with allocated memory is that its size has to be known or possible to estimate before calling the user's solution. It's very often the case, but sometimes such estimation is not possible or easy. There are ways to work around this problem and work with memory allocated by the caller even when its size is not known upfront, but they are out of the scope of this article. In such cases, kata can use memory allocated by the user.
```

content/languages/c/authoring/memory-management-techniques.md
```diff
@@ -6,19 +6,378 @@ sidebar: "language:c"
...
+
+
+### Mixed approach: `malloc` in the solution and `free` in tests
+
+In a vast majority of cases when a kata requires the solution to allocate memory, authors choose the naive approach of allocating the memory in the solution, and releasing it with `free` in the test suite after performing all necessary assertions. This mimics the behavior known from high-level languages where returning an array or object from inside of the user's solution is perfectly valid, but it's not always the best, or even correct, way of working with unmanaged memory in C.
```
DonaldKellett, line 163:
```suggestion
In the vast majority of cases when a kata requires the solution to allocate memory, authors choose the naive approach of allocating the memory in the solution, and releasing it with `free` in the test suite after performing all necessary assertions. This mimics the behavior known from high-level languages where returning an array or object from inside of the user's solution is perfectly valid, but it's not always the best, or even correct, way of working with unmanaged memory in C.
```

DonaldKellett (changes requested):

DonaldKellett (approved):

kazk (approved):

PR #235, opened 2021-01-13
Redone note on memory management

hobovsky:
Maybe having more explicit note in the C tutorial will make the #230 obsolete, unnecessary, or will allow for reducing it?

### DIFF
--- a/content/languages/c/authoring/index.md
+++ b/content/languages/c/authoring/index.md
@@ -68,15 +68,16 @@ Not as much of a problem for C as it is for C++, but still, C authors often forg
 Compiler options related to warnings used by the C runner are somewhat strict and require some discipline to get the code to compile cleanly. `-Wall` and `-Wextra` may cause numerous warnings and some of them are very pedantic. However, code of C kata should still compile cleanly, without any warnings logged to the console. Even when a warning does not cause any problem with tests, users get distracted by them and blame them for failing tests.
 
 
-### Memory management
+### Working with pointers and memory management
 
 Unlike many modern, high-level languages, C does not manage memory automatically. Manual memory management is a very vast and complex topic, with many possible ways of achieving the goal depending on a specific case, caveats, and pitfalls.
 
-Whenever a kata needs to return a string or an array, C authors tend to use the naive technique of allocating the memory in the solution function, and freeing it in the test suite. This approach mimics the behavior known from other languages where returning an array or object from inside of the user's solution is perfectly valid, but it's hardly ever a valid way of working with unmanaged memory.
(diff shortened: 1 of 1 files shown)


error256 (approved):

content/languages/c/authoring/index.md
```diff
@@ -68,15 +68,23 @@ Not as much of a problem for C as it is for C++, but still, C authors often forg
...
 
 
-### Memory management
+### Working with pointers and memory management
 
 Unlike many modern, high-level languages, C does not manage memory automatically. Manual memory management is a very vast and complex topic, with many possible ways of achieving the goal depending on a specific case, caveats, and pitfalls.
 
-Whenever a kata needs to return a string or an array, C authors tend to use the naive technique of allocating the memory in the solution function, and freeing it in the test suite. This approach mimics the behavior known from other languages where returning an array or object from inside of the user's solution is perfectly valid, but it's hardly ever a valid way of working with unmanaged memory.
+Data hidden behind pointers can be arranged in many possible ways. Whenever a kata passes in a pointer to the user's solution, or requires it to return or manipulate a pointer or data referenced by a pointer, it should **explicitly** and **clearly** provide all information necessary to carry out the operation correctly. The information can be put in one or more of following places:
```
Steffan153, line 75:
```suggestion
Data hidden behind pointers can be arranged in many possible ways. Whenever a kata passes in a pointer to the user's solution or requires it to return or manipulate a pointer or data referenced by a pointer, it should **explicitly** and **clearly** provide all information necessary to carry out the operation correctly. The information can be put in one or more of the following places:
```

content/languages/c/authoring/index.md
```diff
@@ -68,15 +68,23 @@ Not as much of a problem for C as it is for C++, but still, C authors often forg
...
+
+- The code itself. Specifying that a pointer points to `const` data can serve as a hint that it has not been allocated dynamically and won't be freed. Size hints for array parameters can help understanding how arays are organized, etc. Correctly specified types can be very helpful, but not always sufficient.
```
Steffan153, line 77:
```suggestion
- The code itself. Specifying that a pointer points to `const` data can serve as a hint that it has not been allocated dynamically and won't be freed. Size hints for array parameters can help understanding how arrays are organized, etc. Correctly specified types can be very helpful, but not always sufficient.
```

content/languages/c/authoring/index.md
```diff
@@ -68,15 +68,23 @@ Not as much of a problem for C as it is for C++, but still, C authors often forg
...
+- [Language-specific paragraph](#description) in the kata description.
+- As a comment in the "Solution setup" snippet.
+- When necessary, [sample tests](/authoring/guidelines/sample-tests/) should present an example of how data is composed, passed to the user solution, fetched from it, worked on, and cleaned up afterwards.
+
+When the structure, layout, or allocation scheme of a pointed data is not described, users cannot know how to implement requirements without causing either a crash or a memory leak. Authors can choose the ownership strategy their kata should use, and the memory can be managed either by the test suite, by the user, or both. However they should be aware what are the advantages and disadvantages of each such strategy, and when and which applies the best.
```
Steffan153, line 82:
```suggestion
When the structure, layout, or allocation scheme of pointed data is not described, users cannot know how to implement requirements without causing either a crash or a memory leak. Authors can choose the ownership strategy their kata should use, and the memory can be managed either by the test suite, by the user, or both. However, they should be aware of the advantages and disadvantages of each such strategy, and when and which applies the best. 
```

Steffan153 (changes requested):

Steffan153 (approved):

ggorlen (approved):

PR #239, opened 2021-01-14
Update moderation.md

Chrono79:
Fixed a couple of typos

### DIFF
--- a/content/community/moderation.md
+++ b/content/community/moderation.md
@@ -36,7 +36,7 @@ They can be recognized by the symbol of a crown (👑) displayed next to their u
 
 ## Moderators
 
-Moderators are a small team of active users selected by admins and are responsible for supervising users's activities. They have earned the trust of the admins from years of contributions and are given access to moderation tools and extra privileges.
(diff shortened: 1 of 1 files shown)


content/community/moderation.md
```diff
@@ -36,7 +36,7 @@ They can be recognized by the symbol of a crown (👑) displayed next to their u
 
 ## Moderators
 
-Moderators are a small team of active users selected by admins and are responsible for supervising users' activities. They have earned the trust of the admins from years of contributions and are given access to moderation tools and extra privileges.
+Moderators are a small team of active users selected by admins and are responsible for supervising users's activities. They have earned the trust of the admins from years of contributions and are given access to moderation tools and extra privileges.
```
kazk, line 39: This is actually correct. The possessive of a plural noun is formed by adding only an apostrophe when the noun ends in `s`.
```suggestion
Moderators are a small team of active users selected by admins and are responsible for supervising users' activities. They have earned the trust of the admins from years of contributions and are given access to moderation tools and extra privileges.
```

Issue #215, opened 2020-12-30
Create a guide for authoring in Ruby


Steffan153:
Create an article similar to https://docs.codewars.com/languages/python/authoring/ but for Ruby.

hobovsky:
@Steffan153 is there anything to do on #252? I think the example test suite is missing?

Steffan153:
Yep that's all that's left.

(Looks like I wasn't noticing all these notifications.)

hobovsky:
You need help finishing it?

Steffan153:
No, I can do it.

PR #252, opened 2021-01-24
Ruby: Authoring kata

Steffan153:
Discussion: #215
Deploy preview: https://deploy-preview-252--reverent-edison-2864ea.netlify.app/languages/ruby/authoring/


content/languages/ruby/authoring.md
```diff
@@ -0,0 +1,156 @@
...
+### Testing framework
+
+Ruby kata should use the [Codewars Ruby testing framework](/languages/ruby/codewars-test/) to implement and execute tests.
+
+<!-- You should read its reference page to find out how to use `describe` and `it` blocks for [organization and grouping](/languages/ruby/codewars-test/#organization-of-tests), what [assertions](/languages/ruby/codewars-test/#assertions) are available, etc. -->
+<!-- TODO above, as the reference page is not done. -->
+
+#### Dynamically generated test cases
+
+It's possible to put functions decorated with `it` in a loop and use them as a construct similar to parameterized test cases known from other testing frameworks, for example:
```
hobovsky, line 73: Are these also called "decorated" in Ruby?

content/languages/ruby/authoring.md
```diff
@@ -0,0 +1,156 @@
...
+### Calling assertions
+
+The Ruby testing framework provides a set of useful [assertions](/languages/ruby/codewars-test/#assertions), but when used incorrectly, they can cause a series of problems:
+- Use of an assertion not suitable for the given case may lead to incorrect test results.
+- Incorrectly used assertions may produce confusing or unhelpful messages.
+
+To avoid the above problems, calls to assertion functions should respect the following rules:
+- The expected value should be calculated _before_ invoking an assertion. The `expected` parameter passed to the assertion should not be a lambda/`Proc`/block, but a value calculated directly beforehand.
+- Appropriate assertion functions should be used for each given test. `Test.assert_equals` is not suitable in all situations. Use `Test.expect` for tests on boolean values, and `Test.expect_error` to test error handling.
+<!-- Use `Test.assert_approx_equals` for floating-point comparisons  -->
```
hobovsky, line 146: I would explicitly mention that there's no assertion for approx equality, because it's somewhat unexpected IMO. Every sane assertions library has it. Snippet is a great idea, just mention it here and link to it.

content/languages/ruby/authoring.md
Steffan153: Oops, that's what happens when you copy-paste xD
```suggestion
It's possible to put `it` blocks in a loop and use them as a construct similar to parameterized test cases known from other testing frameworks, for example:
```

DonaldKellett (approved):

content/languages/ruby/authoring.md
```diff
@@ -0,0 +1,156 @@
+---
+kind: tutorial
+languages: [ruby]
+sidebar: "language:ruby"
+prev: /languages/ruby/codewars-test/
+---
+
+# Ruby: creating and translating a kata
+
+This article is intended for kata authors and translators who would like to create new content in JavaScript. It attempts to explain how to create and organize things in a way conforming to [authoring guidelines](/authoring/guidelines/), shows the most common pitfalls, and how to avoid them.
```
error256, line 10: "in JavaScript"?

error256 (changes requested):

content/languages/ruby/authoring.md
Steffan153:
```suggestion
This article is intended for kata authors and translators who would like to create new content in Ruby. It attempts to explain how to create and organize things in a way conforming to [authoring guidelines](/authoring/guidelines/), shows the most common pitfalls, and how to avoid them.
```

error256 (approved):

content/languages/ruby/authoring.md
```diff
@@ -0,0 +1,156 @@
...
+~~~
+```
+
+## Tasks and Requirements
+
+Some concepts don't always translate well to or from Ruby. Because of this, some constructs should be avoided and some translations just shouldn't be done.
+- Avoid returning different data types depending on the situation (_"Return the result, or the string 'Error' if no result can be found..."_). Ruby is dynamically typed, which is not the case for some other languages. Returning `nil` might be appropriate in some situations, but throwing an error might be better in others.
+
+Some kata just should not be translated into Ruby because it can be difficult to keep their initial idea:
+- Some kata might be meant for another language specifically. For example, a kata about `Python: Learning Classes` should probably not be translated to Ruby, even though Ruby has `class` syntax.
```
Blind4Basics, line 52: I don't feel like it's evident. I practice both (well, not much ruby, yeah, but still) and I honestly don't see why thsi should be avoided in ruby => ?

content/languages/ruby/authoring.md
```diff
@@ -0,0 +1,156 @@
...
+- One somewhat distinctive feature of Ruby assertions is that by default, a failed assertion does not cause a test case to fail early. It can lead to unexpected crashes when an actual value had already been asserted to be invalid, but the execution of the current test case was not stopped and following assertions continue to refer to it.
+- To avoid unexpected crashes in tests, it's recommended to perform some additional assertions before assuming that the answer returned by the user solution has some particular type, form, or value. For example, if the test suite sorts the returned array to verify its correctness, an explicit assertion should be added to check whether the returned object is really a list, and not, for example, `nil`.
+
+## Example test suite
+
+Below you can find an example test suite that covers most of the common scenarios mentioned in this article. Note that it does not present all possible techniques, so actual test suites can use a different structure, as long as they keep to established conventions and do not violate authoring guidelines.
+
+<!-- TODO -->
```
Blind4Basics, line 156: about the example test suite, ruby is a total mess because files are concatenated (preloaded/solution/tests) so everything is in the same scope.
IIRC, it's even possible to access subfunctions, no? I don't actually remember (that mess is part of the reasons I totally dropped ruby), but I'm sure that making the ref solution inaccessible to the user is a hassle with ruby...

content/languages/ruby/authoring.md
Steffan153: You probably could, but you'd want to change the kata title at least, right?

Blind4Basics: oh... But it's only about the title, I feel the reason is out of place, then (but maybe it's just me being oriented by what I knew will follow...?)

content/languages/ruby/authoring.md
Steffan153: I mentioned that on line 133. You just have to use something other than `def`.

content/languages/ruby/authoring.md
Steffan153: I think I just need a better example of something that shouldn't... any ideas?

hobovsky: maybe something what does not translate well into hashes, or duck typing?

Steffan153: You can technically do duck typing in Ruby...
Maybe like something about JS coercion? Ruby doesn't have coercion (yay).

hobovsky: Damn I am so noob in Ruby I have no idea what could apply here.

Maybe dudner methods? Coercion sounds good too.


 Maybe just leave this out?
 
 
I KNOW! Performance kata do not translate well into Ruby! :D

Steffan153: Dunder methods sounds right.

content/languages/ruby/authoring.md
```diff
@@ -0,0 +1,156 @@
...
+~~~
+```
+
+## Tasks and Requirements
+
+Some concepts don't always translate well to or from Ruby. Because of this, some constructs should be avoided and some translations just shouldn't be done.
+- Avoid returning different data types depending on the situation (_"Return the result, or the string 'Error' if no result can be found..."_). Ruby is dynamically typed, which is not the case for some other languages. Returning `nil` might be appropriate in some situations, but throwing an error might be better in others.
+
+Some kata just should not be translated into Ruby because it can be difficult to keep their initial idea:
+- Some kata might be meant for another language specifically. For example, a kata about `Python: Dunder Methods` should probably not be translated to Ruby, as Ruby doesn't have dunder methods.
```
hobovsky, line 52:
```suggestion
- Some kata might be meant for another language specifically. For example, a kata about `Python: Dunder Methods` should probably not be translated to Ruby, as Ruby doesn't have dunder methods, even though similar goal can be achieved with operator overloading and such.
```

I think it will be nice to show more precisely why such translations do not fit.

content/languages/ruby/authoring.md
Blind4Basics: don't ruby classes have `def coerce(self,other)` mehtods? x) 

even dunder methods are a bit edgy because some exist in ruby `=, +, ...`

maybe step out of python, for this? like "learn the difference int/long" in java/C/... doesn't translate to ruby.

?

content/languages/ruby/authoring.md
Blind4Basics: yeah, which leads to order definition troubles x) (assuming you're building lambdas)

DonaldKellett (approved):

content/languages/ruby/authoring.md
kazk: We can add the snippet to the test framework (https://github.com/codewars/ruby-test-framework) that will be used for Ruby 3.0 support. I'm also considering to use RSpec under the hood somehow, then eventually switch to it completely like JS.

hobovsky (approved):
Looks good to me, but I am no Ruby expert so maybe someone else could also take a look.

kazk (approved):
Looks good to me. Thanks for writing this.

By the way, I'm considering to replace the custom test framework with the built-in `minitest` from Ruby 3.0. I'm still experimenting, but it should work the same way except for assertions, so this will be still relevant.

PR #258, opened 2021-01-29
Update glossary.yml

Blind4Basics:
update for : 
https://docs.codewars.com/glossary/#Power-user

### DIFF
--- a/data/glossary.yml
+++ b/data/glossary.yml
@@ -244,6 +244,11 @@
 
     It's useful for [debugging - TODO]() too.
 
+- term: Power user
+  description: |
+    Codewars' users gain honor in different ways. Doing so, they progressively unlock some [privilegees](references/gamification/privileges), hence "powers".
+  acronym: PU
+
 - term: Preloaded (code/section)
   description: |
     In a Kata, the preloaded section is code that is executed _before_ the solution of the user is defined. The code hold by this section isn't visible to the user but every function or variable defined there is accessible to him drugint the executions of his code.


data/glossary.yml
```diff
@@ -244,6 +244,11 @@
 
     It's useful for [debugging - TODO]() too.
 
+- term: Power user
+  description: |
+    Codewars' users gain honor in different ways. Doing so, they progressively unlock some [privilegees](references/gamification/privileges), hence "powers".
```
DonaldKellett, line 249:
```suggestion
Codewars' users gain honor in different ways. Doing so, they progressively unlock some [privileges](references/gamification/privileges), hence "powers".
```

data/glossary.yml
```diff
@@ -244,6 +244,11 @@
...
+  acronym: PU
+
 - term: Preloaded (code/section)
   description: |
     In a Kata, the preloaded section is code that is executed _before_ the solution of the user is defined. The code hold by this section isn't visible to the user but every function or variable defined there is accessible to him drugint the executions of his code.
```
DonaldKellett, line 254:
```suggestion
In a Kata, the preloaded section is code that is executed _before_ the user's solution. The code in this section isn't visible to the user but every function or variable defined there is accessible to the solution during execution.
```

DonaldKellett (changes requested):

DonaldKellett (approved):
I didn't realize that my edits removed the leading whitespace from the affected lines. Perhaps you may want to add them back, but other than that, LGTM

glossary.yml
```diff
@@ -244,9 +244,14 @@
 
     It's useful for [debugging - TODO]() too.
 
+- term: Power user
```
kazk, line 247: Power user is not about having "powers".


> A power user is a user of computers, software and other electronic devices, who uses advanced features of computer hardware, operating systems, programs, or websites which are not used by the average user. A power user might not have extensive technical knowledge of the systems they use but is rather characterized by competence or desire to make the most intensive use of computer programs or systems.
>
> https://en.wikipedia.org/wiki/Power_user

Blind4Basics: Mmmh, I went too much "litteral" here (happy frenglish... x) ). What about:

> Codewars' users gain honor in different ways. Doing so, they progressively unlock some [privileges](references/gamification/privileges) giving them access to extra features like the possibility to approve kata, maintain kata they didn't author, resolve comments, ...

?

kazk: It's not about gaining privileges. If they don't actively make use of them, they're not power users.

I think the key point in the Wikipedia is:

> **desire** to make the most intensive use

---

The opposite of power user is average user. Those users that only uses Codewars to train are average users even if they have many privileges unlocked. They don't use features like authoring kata, or get involved with maintaining Codewars.

Users' activities can be broken down like the following:

1. Train
2. Review
3. Create/Translate
4. Maintain

Probably 99% of the users only does 1 and partially 2 (solving beta). 3 and 4 are power users that affects Codewars.

Issue #266, opened 2021-02-12
python dynamic test registration: potential to fail to register unique tests


namedots:
I was asked by hobovsky to describe this situation -

There's an example [here](https://docs.codewars.com/languages/python/authoring/#dynamically-generated-test-cases):

```python
@test.describe("Generated test cases")
def tests_with_generated_test_cases()
    test_cases = generate_test_cases()
    for msg, input, expected in test_case:
        @test.it(msg)
        def _():
            actual = user_solution(input)
            test.assert_equals(actual, expected)
```

which has a potential bug in it - the test framework could decide to call `_` after the loop has finished
what this will do is that all instances of `_` will have the same value for `input` and `expected` as those are variables existing in the surrounding scope and python closures do not capture current values but instead the variables themselves

as a small demonstration:

```python
funs = []

# register 0 1 2 3
for i in range(4):
    def f():
        print(i)
    funs.append(f)

for f in funs:
    f()  # expected 0 1 2 3
         # but prints 3 3 3 3
```

this pattern is totally fine with the current test framework which immediately calls the registered function and maybe it's reasonable. still, the potential is there and this does happen in javascript - node 8.something calls the functions immediately, while a more recent version holds on to the callbacks to be invoked later (and javascript has the same capturing behaviour for closures) sorry I don't have reproducing code for that, so that part is anecdotal evidence

Personally, I've been doing this:

```python
def check(indata, expected):
    reel, spin = indata
    test.it(f"fruit({reel!r}, {spin!r})")(
        lambda: test.assert_equals(fruit(reel, spin), expected)
    )

# ...

for indata, expected in test_cases:
    check(indata, expected)
```

no clue what other people think of that but it gets around the problem since it immediately binds the test parameters

kazk:
> the test framework could decide to call `_` after the loop has finished

This won't happen because it's a huge breaking change. This will completely changes the test framework. I'd rather throw it away and use a proper test framework :P

> (and javascript has the same capturing behaviour for closures)

JavaScript have block scope (`let`/`const`) (or you can also use IIFE):

```javascript
// All of these will work
{
  const funs = [];
  for (let i = 0; i < 5; ++i) funs.push(() => i);
  funs.forEach(f => console.log(f()));
}
{
  const funs = [];
  for (const i of [0,1,2,3,4]) funs.push(() => i);
  funs.forEach(f => console.log(f()));
}
{ // If you "really" need "var" for some strange reason, use IIFE
  const funs = [];
[...]

PR #268, opened 2021-02-18
Notes on variables captured in test closures

hobovsky:
Motivated by #266 

Preview: [for Python](https://deploy-preview-268--reverent-edison-2864ea.netlify.app/languages/python/codewars-test/#organization-of-tests), [for JavaScript](https://deploy-preview-268--reverent-edison-2864ea.netlify.app/languages/javascript/authoring/#dynamically-generated-test-cases).

### DIFF
--- a/content/languages/javascript/authoring.md
+++ b/content/languages/javascript/authoring.md
@@ -103,7 +103,7 @@ describe("Generated test cases", () => {
 });
 ```
 
-This technique is liked by authors familiar with testing frameworks that provide parameterized or generated test cases out of the box, like NUnit or JUnit. However, some caution is needed when this approach is used. Test suites organized like this can become large and can flood the test output panel with many entries, making it unreadable or causing performance problems in client browsers. Additionally, to avoid problems with captured variables, declare them with `const` or `let`, and never use `var` or global variables.
+This technique is liked by authors familiar with testing frameworks that provide parameterized or generated test cases out of the box, like NUnit or JUnit. However, some caution is needed when this approach is used. Test suites organized like this can become large and can flood the test output panel with many entries, making it unreadable or causing performance problems in client browsers.
 
 
 ### Chai
--- a/content/languages/javascript/authoring.md
+++ b/content/languages/javascript/authoring.md
@@ -103,7 +103,8 @@ describe("Generated test cases", () => {
 });
 ```
 
-This technique is liked by authors familiar with testing frameworks that provide parameterized or generated test cases out of the box, like NUnit or JUnit. However, some caution is needed when this approach is used. Test suites organized like this can become large and can flood the test output panel with many entries, making it unreadable or causing performance problems in client browsers.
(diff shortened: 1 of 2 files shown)


content/languages/javascript/authoring.md
```diff
@@ -103,7 +103,8 @@ describe("Generated test cases", () => {
 });
 ```
 
-This technique is liked by authors familiar with testing frameworks that provide parameterized or generated test cases out of the box, like NUnit or JUnit. However, some caution is needed when this approach is used. Test suites organized like this can become large and can flood the test output panel with many entries, making it unreadable or causing performance problems in client browsers.
+This technique is liked by authors familiar with testing frameworks that provide parameterized or generated test cases out of the box, like NUnit or JUnit. However, some caution is needed when this approach is used. Test suites organized like this can become large and can flood the test output panel with many entries, making it unreadable or causing performance problems in client browsers. Additionally, to avoid problems with captured variables, declare them with `const` or `let`, and never use `var` or global variables.
```
kazk, line 106: It's fine as long as the value doesn't "leak" and get overwritten. So `forEach` (like the example above) is fine as well. Immutable globals shouldn't cause issues either.

content/languages/python/codewars-test/index.md
```diff
@@ -41,6 +41,8 @@ This is a convention used in many Python kata, but it's not a requirement, and a
 
 Tests in the Python testing framework are composed of functions decorated with a set of Python decorators. All such functions are automatically discovered and run. The final result for every test block is determined by contained assertions and reported along with its measured execution time.
 
+Test decorators are created in a way that decorated functions are immediatelly invoked and executed. As a result, they capture and use the current values of variables from enclosing scope.
```
kazk, line 44: Sounds confusing.

```suggestion
Test decorators call the functions immediately, so it's safe to use them in a loop.
```

---

> Tests in the Python testing framework are composed of functions decorated with a set of Python decorators.

It's confusing because it's not really "decorating". It's more of a syntax sugar/hack to call the function immediately within a wrapper. I don't think we should go into the details in this document meant for general users, so mentioning this is safe should be enough.

DonaldKellett (approved):

content/languages/javascript/authoring.md
hobovsky: I have no good idea how to word it, mostly because I think I am too noob with JS and I do not fully understand the problem. Could you please suggest something?

kazk: See https://stackoverflow.com/a/750506 for a list of solutions and https://2ality.com/2015/02/es6-scoping.html scoping details.

To be honest, I'm not sure if we should include this in our docs anymore. JavaScript has gotchas like most languages, and we can't cover all. Modern JS has `let`/`const` and `var` should be avoided anyway.

The Python decorator is definitely worth mentioning because it's really weird.

kazk, line 106:
```suggestion
This technique is liked by authors familiar with testing frameworks that provide parameterized or generated test cases out of the box, like NUnit or JUnit. However, some caution is needed when this approach is used. Test suites organized like this can become large and can flood the test output panel with many entries, making it unreadable or causing performance problems in client browsers.
```

Issue #242, opened 2021-01-16
Working with floating point numbers


hobovsky:
Create a tutorial explaining why floating point numbers are difficult to work with, what problems they cause in kata, and how to avoid pitfalls related to them when authoring a kata.

Maybe the tutorial might be not limited only to authoring, but could be made useful also for users encountering problems while training?

- precision loss, non-associativity, conversions
- comparison, assertions 
- formatting (also in assertion messages)
- alternatives
- rounding, and why not rounding. Rounding modes

hobovsky:
I've decided to hide the "why" section and I will probably remove it, because it seems to introduce too many details. But still I do not like the fact that it will probably strip the article from its educational aspect.

@kazk could you hint some good location for the article files in the repo?

All reviews are appreciated, and any remarks are welcome.

kazk:
Under `/authoring/kata/` like you have should be fine for now.

imjasonmiller:
I would definitely appreciate some code examples for authors regarding assertions. While I've used `approx` myself for approximate float comparison, there are a few others crates that seem to do the same in Rust:

* [approx](https://github.com/brendanzab/approx)
* [float-cmp](https://github.com/mikedilger/float-cmp)
* [float_eq](https://github.com/jtempest/float_eq-rs)

The author of `float_eq`, the newest crate, [expanded on the differences](https://www.reddit.com/r/rust/comments/fzwxiy/announcing_float_eq_for_explicitly_bounded/) between it and the other listed crates.

I'd be interested in hearing what others have to say, perhaps @error256?

kazk:
> The [`approx`](https://lib.rs/crates/approx) and [`float-cmp`](https://lib.rs/crates/float-cmp) crates provide a similar style of general comparison operations, whereas [`assert_float_eq`](https://lib.rs/crates/assert_float_eq) focuses specifically on assertions. The [`almost`](https://lib.rs/crates/almost) crate instead divides its API into algorithms comparing against zero and non-zero values. In contrast, [`efloat`](https://lib.rs/crates/efloat) takes the approach of tracking the error bounds of values as operations are applied.
>
> https://github.com/jtempest/float_eq-rs#related-efforts

hobovsky:
I'd be reluctant to put too much language-specific information in this article. I tried to keep it to the minimum, and mention only some notable exceptions.

[...]

PR #269, opened 2021-02-18
Tutorial: Working with floating-point numbers

hobovsky:
Preview: [here](https://deploy-preview-269--reverent-edison-2864ea.netlify.app/authoring/kata/floating-point/)
Discussion: #242


content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,277 @@
...
+```
+
+You need to remember that whenever you print or format floating-point values as strings, they should always be formatted with at least as many significant digits as the tolerance used for comparisons.
+
+
+### Use alternatives
+
+There are some ways to mitigate some issues related to floating-point numbers by using some other means, language constructs, classes, libraries, etc. Unfortunately, they are usually language-specific and can turn out to be problematic in kata, because translating kata between languages can become difficult. However, some possibilities are:
+
+- You can use arbitrary-precision decimal types or similar, for example `java.math.BigDecimal` in Java, `decimal` (floating-point decimal primitive type) in C#, `decimal` module in Python, etc.
```
error256, line 276: > some ways to mitigate some issues related to floating-point numbers

Using floating point numbers to mitigate issues with floating point numbers? Arbitrary precision is still finite, so it's just binary vs decimal here, which makes the difference only if the input is a finite decimal.

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,220 @@
...
+To find out what assertions are appropriate for floating-point comparisons in your language you should go through the documentation of the testing framework you use. For example, for JavaScript it's `chai.assert.closeTo`, and for Python it's `codewars_test.assert_approx_equals`.  
+Some testing frameworks used on Codewars unfortunately lack proper assertions. This is the case for example for C++, or Ruby. In such case, you need to provide your own fuzzy comparison function. It's a difficult task to do it correctly though, so don't try to create one on your own. There's plenty of materials on Internet which will help you with this task, for example [this great guide](https://floating-point-gui.de/) and [its page](https://floating-point-gui.de/errors/comparison/) dedicated strictly for floating point comparisons.
```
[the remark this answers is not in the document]
hobovsky: Uh, totally slipped my mind. Thanks for spotting.

content/authoring/kata/floating-point.md
hobovsky: I was wondering how/if to differentiate between decimal fixed-scale numbers (`java.math.BigDecimal`), floating-point decimal numbers, (`decimal` in C#), all of them as opposed to floating-point binary numbers (IEEE-754). So basically the sentence would mean something like _"Using decimal floating point numbers to mitigate issues with IEEE-754 floating point numbers"_ . Or something like that.

I am not sure about the _" it's just binary vs decimal"_, or maybe I am not sure about a proper nomenclature, because some nitpicker could argue that C# `decimal` is still binary (IIRC it's encoded in some form of BCD).

But yeah, maybe just "use decimal data type if appropriate" would be enough, I am just afraid that someone could get an impression that decimal data types from different libraries/languages are functionally equivalent.

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+Above program prints:
+
+```text
+Is the area equal to 1.21? 
+No, it's 1.2100000000000002
+```
+
+That's because variable `a` has never been equal to `1.1` to begin with! When you look at the code it might look like that, but compiler silently converted literal value `1.1` to the nearest representable floating-point value and stored `1.100000000000000088817841970012523233890533447265625` in variable `a`.
```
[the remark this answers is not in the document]
hobovsky: Uh, I was to remove the backticks around the numbers and forgot.

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,220 @@
...
+
+- They can hold fractional values, like `25.5` or `-7.125`. Fractions can be as small as ~1.7\*10<sup>-308</sup>
+- They have much wider range than integral types of the same size. Signed 64-bit integer data type can hold values up to ~9\*10<sup>18</sup>, while 64-bit floating-point value can go up to ~1.7\*10<sup>308</sup>.
+
+However, to achieve these goals, floating-point values trade off their _precision_, what more or less means that there are some values which cannot be represent exactly. They are capable of storing up to 52 bits, or approximately 15 decimal digits, and anything beyond this is lost. Values which would require better precision are internally rounded and stored just as approximations. And this leads to many problems which often can come up as surprising.
+
+### Representability
+
+As it was already mentioned, precision of a floating-point value is limited to 52 bits, or approx. 15 decimal digits. This means that some values simply cannot be stored in a floating-point variable. For example, value `10000000000000000` (or 1*10<sup>16</sup>) can be represented exactly, but there's no `10000000000000001`! How do you think, what will following code print?
```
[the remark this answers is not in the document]
hobovsky: I will think of some better example if we decide this part to stay.

Do you think it's worth to keep the _"Strange things about floating-point values"_ part, or just remove it and keep only _"How to avoid problems with floating-point numbers?"_

content/authoring/kata/floating-point.md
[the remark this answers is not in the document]
kazk: > In such case, you need to provide your own fuzzy comparison function.

For those we maintain, I'd appreciate contributions instead. Ruby test framework to be used in the next version update is open sourced.

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,277 @@
...
+
+One of such problems is overflow in Javascript. For example, consider a kata with following task: _"Given two natural numbers `a` and `b`, calculate and return their least common multiple."_ Random tests are careful enough to generate such values of `a` and `b`, which always should give the resulting LCM as less than `Number.MAX_SAFE_INTEGER`. However, the reference solution uses following implementation:
+
+```javascript
+function lcm(a, b) {
+    return a * b / gcd(a, b); //gcd is a helper function to calculate greatest common divisor
+}
+```
+
+Can you see where potential problem is? Even though `a`, `b`, and final result are guaranteed to be less than `Number.MAX_SAFE_INTEGER`, intermediate value of `a * b` can overflow, resulting in incorrect result being returned. Now reference solution has a bug!
```
DonaldKellett, line 194:
```suggestion
Can you see where the problem is? Even though `a`, `b`, and final result are guaranteed to be less than `Number.MAX_SAFE_INTEGER`, the intermediate value of `a * b` can overflow, resulting in an incorrect result being returned. Now the reference solution has a bug!
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,277 @@
...
+
+You need to make sure that your reference solution is correct and can handle all test inputs which will be generated and fed to it. Otherwise, correctness of tests depends on the favor of random number generator, and some valid solutions might fail the tests in an unpredictable, and difficult to debug way.
```
DonaldKellett, line 196:
```suggestion
You need to make sure that your reference solution is correct and can handle all test inputs which will be generated and fed to it. Otherwise, the correctness of tests depends on the random number generator, and some valid solutions may fail the tests in an unpredictable manner.
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,277 @@
...
+
+#### Division
+
+In languages like JavaScript or Python, floating-point values can be introduced by mistake when performing division. JavaScript has no integer division operator, and operation of `a / b` always results in floating-point values. If you are interested in integer division, you need to remember to convert the result to integer by yourself. Python has two division operators: `a // b` for integer division (or, "real floor division"), and `a / b` for "real" division. Both operators can be easily confused, and `/` will convert integers to floats and return a float.
```
DonaldKellett, line 200:
```suggestion
In languages like JavaScript or Python, floating-point values can be introduced by mistake when performing division. JavaScript has no integer division operator, so `a / b` always results in a floating-point value. If you are interested in integer division, you need to remember to convert the result to an integer by yourself. Python has two division operators: `a // b` for integer division (or, "real floor division"), and `a / b` for "real" division. Both operators can be easily confused, and `/` will convert integers to floats and return a float.
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,277 @@
...
+
+
+### Use relative comparisons and avoid strict equality
+
+Have you ever encountered a problem that some solution does not pass because tests expect a slightly different value, even though you used right formula?
```
DonaldKellett, line 205:
```suggestion
Have you ever encountered a problem that some solution does not pass because tests expect a slightly different value, even though you used the right formula?
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,277 @@
...
+
+```text
+Test Results:
+convert_temperature
+    given 2700 (steel melting point)
+        expected 1482.2222222222224 to equal 1482.2222222222222
+```
+
+This usually happens when tests use incorrect assertion and do not account for the fact that user's solution can return values which are not exactly the same as the ones produced by a reference solution, but still correct.
```
DonaldKellett, line 214:
```suggestion
This usually happens when tests use an incorrect assertion method and do not account for the fact that the user's solution can return a correct answer which is not identical to the one produced by the reference solution.
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,277 @@
...
+
+To correctly test for floating-point values, tests should use a thing called _"approximate equality tests"_, _"fuzzy equality"_, or _"comparison with tolerance"_. Basically it boils down to using a dedicated assertion provided by a majority of popular testing frameworks or assertion libraries. Such asertions usually take as parameters an expected value, actual value, and a value of tolerance which tests agree on. Assertion accepts all answers which are equal to the expected result, or do not differ from it more than allowed.
```
DonaldKellett, line 216:
```suggestion
To correctly test for floating-point values, tests should use _"approximate equality tests"_, _"fuzzy equality"_, or _"comparison with tolerance"_. Basically it boils down to using a dedicated assertion provided by the majority of popular testing frameworks or assertion libraries. Such assertions usually take as parameters the expected value, actual value, and a value of tolerance which the tests agree on. The assertion accepts all answers which are equal to the expected result, or do not differ from it beyond the provided tolerance.
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,277 @@
...
+```text
+Test Results:
+sample_test
+    test_basic
+        Expected 1.210000 +/- 1e-9, but was 1.210000
+```
+
+This issue is not limited only to C, other languages are also affected. 
+
+So... what's wrong? Why tests fail, if expected value and actual value are both `1.210000`?
```
DonaldKellett:
```suggestion
So... what's wrong? Why do the tests fail if the expected and actual values are both `1.210000`?
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,277 @@
...
+Most probably, two things are wrong:
+- User solution returns invalid answer, which does not fall into required tolerance, and
```
DonaldKellett, line 249:
```suggestion
- The user's solution returned an invalid answer outside of the required tolerance, and
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,277 @@
...
+- Tests display assertion message incorrectly, truncating some meaningful digits.
```
DonaldKellett, line 250:
```suggestion
- The tests display the assertion message incorrectly, truncating some meaningful digits.
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,277 @@
...
+
+Actually, both numbers are different, and actual answer is wrong by more than allowed tolerance. But when formatting floating-point values, many languages use only some part of digits, and truncate the least meaningful ones, even when they are still important. When you fix formatting of the message, you get better feedback:
```
DonaldKellett, line 252:
```suggestion
Actually, both numbers are different, and the actual answer differs from the expected value by an amount greater than the allowed tolerance. But when formatting floating-point values, most languages only display the first few digits. When you fix the formatting of the message, you get better feedback:
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,277 @@
...
+
+You need to remember that whenever you print or format floating-point values as strings, they should always be formatted with at least as many significant digits as the tolerance used for comparisons.
+
+
+### Use alternatives
+
+There are some ways to mitigate some issues related to floating-point numbers by using some other means, language constructs, classes, libraries, etc. Unfortunately, they are usually language-specific and can turn out to be problematic in kata, because translating kata between languages can become difficult. However, some possibilities are:
+
+- You can use arbitrary-precision decimal types or similar, for example `java.math.BigDecimal` in Java, `decimal` (floating-point decimal primitive type) in C#, `decimal` module in Python, etc.
+- If your kata uses rational numbers, storing numbers as a pair of numerator and denominator can help. Python provides `fractions` module which offers such types out of the box, and Haskell has `Rational`.
```
DonaldKellett:
```suggestion
- If your kata uses rational numbers, storing numbers as a pair of integers (numerator, denominator) can help. Python provides a `fractions` module which offers such types out of the box, and Haskell has `Rational`.
```

DonaldKellett (changes requested):

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,279 @@
...
+## What are floating-point numbers?
+
+Floating-point numbers come in many different shapes and colors, but one (or, if you prefer, two) of them is the most widespread, the most common, and, at the same time, the most problematic for newbies (and not only). Whenever you see or use anything called _"float"_, _"double"_, _"single-"_ or _"double-precision floating point number"_, _"IEEE-754 number"_, you need to be careful, because things can often turn out surprising and not as easy as one might think. Through the rest of the article, the term _"floating-point numbers"_ will refer specifically to **IEEE-754 double-precision values**, unless explicitly stated otherwise.
+
+Floating-point numbers are usually not specific to any particular language, system, or architecture. IEEE-754 numbers are widely adopted and used across many platforms, languages, and compilers. With some small exceptions, anything written in this article applies in the same way to Python, C, JavaScript, or Haskell.
+
+<details>
+
+<summary>
+"Why"
```
kazk, line 23: Please don't do this. `<summary>` should be descriptive and shouldn't contain anything else like this note block.

kazk (changes requested):

content/authoring/kata/floating-point.md
hobovsky: I used this only to collapse a section which I am not sure about. I am not sure if it is educational enough and should stay, or is it too detailed and should be removed.

If you guys think it should be removed, I will remove it. If you think it should stay, I will move it out of `<details>`. Or maybe you have some other idea?

hobovsky: Unless you mean just the _"Why"_ "caption", it can be removed.

```suggestion
```

kazk: No, this:
image

`<summary>` should be short and fit next to that triangle. Spec-wise it should only contain "phrasing content" (something that will work inside `<p>`). Browsers are lax because users don't write correct HTML :(

This huge summary is bad from usability point of view, try reproducing this scroll position

image

then click the `<summary>`. The triangle points down, but it looks like nothing is showing up because it's happening below the block.

hobovsky: I've removed the collapsible thing. Maybe now seeing how long the whole page gets it will be easier for reviewers to suggest whether the "why" part should stay or not.

content/authoring/kata/floating-point.md
hobovsky: I've reworded this part a bit, got rid of unnecessary deails.

content/authoring/kata/floating-point.md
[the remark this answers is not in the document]
error256: IDK, although these things aren't really strange, they are more like possibly unexpected.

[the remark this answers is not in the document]
hobovsky: The "why" part was redone, and greatly simplified. There's no more "strange things" there, and troublesome example above was removed.

I am still not sure whether to keep it, but now it's at least reduced.

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,194 @@
...
+
+
+## What are floating-point numbers?
+
+Floating-point numbers come in many different shapes and colors, but one (or, if you prefer, two) of them is the most widespread, the most common, and, at the same time, the most problematic for newbies (and not only). Whenever you see or use anything called _"float"_, _"double"_, _"single-"_ or _"double-precision floating point number"_, _"IEEE-754 number"_, you need to be careful, because things can often turn out surprising and not as easy as one might think. Through the rest of the article, the term _"floating-point numbers"_ will refer specifically to **IEEE-754 double-precision values**, unless explicitly stated otherwise.
+
+Floating-point numbers are usually not specific to any particular language, system, or architecture. IEEE-754 numbers are widely adopted and used across many platforms, languages, and compilers. With some small exceptions, anything written in this article applies in the same way to Python, C, JavaScript, or Haskell.
+
+
+## Wy are floating-point numbers problematic?
```
DonaldKellett, line 21:
```suggestion
## Why are floating-point numbers problematic?
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,194 @@
...
+
+Technical details related to the inner representation of floating-point values and how they work are out of scope of this article, but the main problem with floating-point numbers in context of Codewars kata boils down to the fact that calculations which are equivalent from mathematical point of view, but implemented differently, can give slightly different results for the same inputs, and all of them are considered correct even if they are not equal one to each other.
```
DonaldKellett, line 23:
```suggestion
Technical details related to the inner representation of floating-point values and how they work are out of scope of this article, but the main problem with floating-point numbers in the context of Codewars kata boils down to the fact that calculations which are equivalent from a mathematical point of view, but implemented differently, can give slightly different results for the same inputs, and all of them are considered correct even if they are not equal to each other.
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,194 @@
...
...t your kata would require rounding to a nearest integer. When tests are run, one of random tests might generate an input which after all calculations would conclude with a result of 13.5, and tests would expect a rounded answer of 14. Now, depending on details of implementation like exact formula, order of operations, used functions and libraries, user ___A___ comes up with the result of 13.50000000001, and user ___B___ arrives to 13.499999999999. Both answers are correct and valid from technical point of view, but after rounding to 14, user ___A___ gets their answer accepted, and the rounded answer of 13 from user ___B___ gets (incorrectly) rejected. Now user ___B___ creates an issue for your kata and complains that their valid solution does not pass tests. You verify tests with your reference solution, see no error, and resolve the issue as _"cannot reproduce"_. A couple of weeks later another user who happened to use the same formula as user ___B___ comes, does not pass the tests, and creates another issue, and so on... There's quite a number of kata on Codewars affected by similar issue, and half of the users attemting them might like them, and another half probably hates them.
```
DonaldKellett, line 92:
```suggestion
For example, let's assume that your kata would require rounding to the nearest integer. When tests are run, a random test might generate an input which after all calculations would conclude with a result of 13.5, and tests would expect a rounded answer of 14. Now, depending on details of implementation like the exact formula, order of operations, used functions and libraries, user ___A___ comes up with the result of 13.50000000001, and user ___B___ arrives to 13.499999999999. Both answers are correct and valid from a technical point of view, but after rounding to 14, user ___A___ gets their answer accepted, and the rounded answer of 13 from user ___B___ gets (incorrectly) rejected. Now user ___B___ creates an issue for your kata and complains that their valid solution does not pass tests. You verify the tests with your reference solution, see no error, and resolve the issue as _"irreproducible"_. A couple of weeks later another user who happened to use the same formula as user ___B___ comes, does not pass the tests, and creates another issue, and so on... There's quite a number of kata on Codewars affected by a similar issue, and half of the users attempting them might like them, while the other half probably hate them.
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,194 @@
...
+
+Conversions by flooring, ceiling, or truncating are affected by the same issue, just for another category of values. User ___A___ might come up with the result of 12.999999999999 before rounding, and user ___B___ with value 13.000000000001, and if tests expect a floored, ceiled, or truncated value, the answer from one user will be accepted, and from the other one - rejected.
```
DonaldKellett, line 94:
```suggestion
Conversions by flooring, ceiling, or truncating are affected by the same issue, just for another category of values. User ___A___ might come up with the result of 12.999999999999 before rounding, and user ___B___ with value 13.000000000001, and if tests expect a floored, ceiled, or truncated value, the answer from one user will be accepted, and from the other one rejected.
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,194 @@
...
+To find out what assertions are appropriate for floating-point comparisons in your language you should go through the documentation of the testing framework you use. For example, for JavaScript it's `chai.assert.closeTo`, and for Python it's `codewars_test.assert_approx_equals`.  
+Some testing frameworks used on Codewars unfortunately lack proper assertions. This is the case for example for Ruby. In such case, a function for fuzzy comparisons has to be provided. It's a difficult task to do it correctly though, so don't try to create one on your own. Request missing functionality on [Codewars code runner](https://github.com/codewars/runner/issues) board and necessary packages or functions will be added.
```
DonaldKellett, line 114:
```suggestion
Some testing frameworks used on Codewars unfortunately lack proper assertions. This is the case for example for Ruby. In such case, a function for fuzzy comparisons has to be provided. It's a difficult task to do it correctly though, so don't try to create one on your own. Request missing functionality on the [Codewars code runner](https://github.com/codewars/runner/issues) board and necessary packages or functions will be added.
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,194 @@
...
+
+
+### Know how floating-point values and operations work in your language
+
+Some languages, especially dynamically typed ones, are particularly susceptible to problems related to floating-point values. For example, JavaScript does not use integers at all, and all arithmetic calculations operate on floating-point numbers. Everything should be good as long as input, output, and intermediate values stay accurate or do not exceed the value of `Number.MAX_SAFE_INTEGER`, but due to some bug this can happen, and when it happens, it can be difficult to spot.
+
+#### Overflow
+
+One such problem is overflow in JavaScript. For example, consider a kata with the following task: _"Given two natural numbers `a` and `b`, calculate and return their least common multiple."_ Random tests are careful enough to generate such values of `a` and `b` which should always give a resulting LCM less than `Number.MAX_SAFE_INTEGER`. However, the reference solution uses following implementation:
```
DonaldKellett, line 123:
```suggestion
One such problem is overflow in JavaScript. For example, consider a kata with the following task: _"Given two natural numbers `a` and `b`, calculate and return their least common multiple."_ Random tests are careful enough to generate such values of `a` and `b` which should always give a resulting LCM less than `Number.MAX_SAFE_INTEGER`. However, the reference solution uses the following implementation:
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,194 @@
...
+
+```javascript
+function lcm(a, b) {
+    return a * b / gcd(a, b); //gcd is a helper function to calculate greatest common divisor
+}
+```
+
+Can you see where the problem is? Even though `a`, `b`, and final result are guaranteed to be less than `Number.MAX_SAFE_INTEGER`, the intermediate value of `a * b` can overflow, resulting in an incorrect result being returned. Now the reference solution has a bug!
```
DonaldKellett, line 131:
```suggestion
Can you see where the problem is? Even though `a`, `b`, and the final result are guaranteed to be less than `Number.MAX_SAFE_INTEGER`, the intermediate value of `a * b` can overflow, resulting in an incorrect result being returned. Now the reference solution has a bug!
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,194 @@
...
+
+You need to remember that whenever you print or format floating-point values as strings, they should always be formatted with at least as many significant digits as the tolerance used for comparisons.
+
+
+### Use alternatives
+
+There are some ways to mitigate some issues related to floating-point numbers by using some other means, language constructs, classes, libraries, etc. Unfortunately, they are usually language-specific and can turn out to be problematic in kata, because translating kata between languages can become difficult. However, some possibilities are:
+
+- If the kata task is related to decimal numbers (for example monetary values), you can use decimal types, for example `java.math.BigDecimal` in Java, `decimal` in C#, `decimal` module in Python, etc.
+- If domain of your kata is limited to rational numbers, storing values as a pair of integers (numerator, denominator) can help. Python provides a `fractions` module which offers such types out of the box, and Haskell has `Rational`.
```
DonaldKellett, line 188:
```suggestion
- If the domain of your kata is limited to rational numbers, storing values as a pair of integers (numerator, denominator) can help. Python provides a `fractions` module which offers such types out of the box, and Haskell has `Rational`.
```

content/authoring/kata/floating-point.md
```diff
@@ -0,0 +1,194 @@
...
+
+You should always consult documentation of your [language](/languages/) and available libraries to know what are possible options.
```
DonaldKellett, line 190:
```suggestion
You should always consult the documentation for your [language](/languages/) and available libraries to know the possible options available.
```

DonaldKellett (changes requested):

hobovsky:
I am going to merge the article because there was no real negative feedback on it, I think it can prove useful, and it already went through a couple of rounds of reviews. Recent commit of Donald's grammar suggestions was not accepted by him, but if anyone finds any language issues (or any other), let me know or just create a PR with a fix.

PR #277, opened 2021-04-14
Use Docusaurus

kazk:
Decided to use [Docusaurus](https://docusaurus.io/) because it's great out of the box, and allows extending Markdown with [MDX](https://mdxjs.com/) like shortcodes in Zola. I still like Zola for simplicity, but it requires more setup, and I don't have the time. Full text search will be added using [DocSearch](https://docsearch.algolia.com/).

The glossary page is not supported at the moment, but it will be.

---

image

image


kazk:
Preview: https://deploy-preview-277--reverent-edison-2864ea.netlify.app/

hobovsky:
Do you need any help here, migrating content or anything?

kazk:
Can you browse the preview and let me know if you find anything obviously broken? There are some things like sidebar labels/ordering and content structures that can be improved, but this PR should focus on the migration. Feel free to open issues about anything to improve, though, so we don't forget.

Docusaurus has link checker, so all links within the docs should be valid.

When this is deployed, I'll [apply](https://docsearch.algolia.com/apply/) for DocSearch, and add the search bar after we get accepted.

kazk:
I think it's good enough to merge. We'll fix minor issues like titles as we go.

PR #287, opened 2021-04-16
Add info on confusing timeouts

hobovsky:
Explain what does it mean when test output panel is all green, but tests time out.

Preview: [this paragraph](https://deploy-preview-287--reverent-edison-2864ea.netlify.app/training/troubleshooting#pass-but-timeout).


content/training/troubleshooting.mdx
```diff
@@ -52,6 +52,10 @@ There is a couple of possible reasons for such behavior:
...
 - If all tests fail with some message similar to `"Expected 'Something' but got None"` or something along the lines (depending on language and assertions library used) it might mean that you only print your answer to the stdout instead of returning it from your function. This problem happens especially often for people who do not understand the difference between printing value and returning it, or when someone comes from some other competitive programming website, which actually requires values to be printed and not returned.
 
+### All tests pass, but I get a time out {#pass-but-timeout}
+
+Sometimes users are confused when their solution is valid, passes all tests, test output panel is green, but they cannot submit the solution because tests time out. Such situatiom means that solution does _not_ pass all the tests. Most likely, the solution returns a valid answer and passes tests up to some point, but it's too slow and then gets forcibly aborted by the runner due to exceeded execution time. Solution needs to be improved and made more performant. Some hints can be found [here](#timeout).
+
```
awesomeAD1, line 58: Some fixes:

```suggestion
Sometimes users are confused when their solution is valid, passes all tests, test output panel is green, but they cannot submit the solution because tests time out. Such a situation means that solution does _not_ pass all the tests. Most likely, the solution returns a valid answer and passes tests up to some point, but it's too slow and then gets forcibly aborted by the runner due to exceeded execution time. Solution needs to be improved and made more performant. Some hints can be found [here](#timeout).
```

content/training/troubleshooting.mdx
```diff
@@ -52,6 +52,10 @@ There is a couple of possible reasons for such behavior:
 - When the answer is expected to be a floating-point number, it might not be precise enough to compare equal with the expected value, but assertion message rounds both so they look the same. Check for possible rounding errors or floating-point accuracy issues.
 - If all tests fail with some message similar to `"Expected 'Something' but got None"` or something along the lines (depending on language and assertions library used) it might mean that you only print your answer to the stdout instead of returning it from your function. This problem happens especially often for people who do not understand the difference between printing value and returning it, or when someone comes from some other competitive programming website, which actually requires values to be printed and not returned.
```
awesomeAD1, line 53: Suggested fixes:

```suggestion
- When the answer is expected to be a floating-point number, it might not be precise enough to compare equal with the expected value, but the assertion message rounds both so they look the same. Check for possible rounding errors or floating-point accuracy issues.
- If all tests fail with some message similar to `"Expected 'Something' but got None"` or something along those lines (depending on language and assertions library used), it might mean that you only print your answer to the stdout instead of returning it from your function. This problem occurs especially often for people who do not understand the difference between printing a value and returning it, or when someone comes from some other competitive programming website, which actually requires values to be printed and not returned.
```

content/training/troubleshooting.mdx
```diff
@@ -49,10 +49,12 @@ There is a couple of possible reasons for such behavior:
...
+- If all tests fail with some message similar to `"Expected 'Something' but got None"` or something along those lines (depending on language and assertions library used), it might mean that you only print your answer to the stdout instead of returning it from your function. This problem occurs especially often for people who do not understand the difference between printing a value and returning it, or when someone comes from some other competitive programming website, which actually requires values to be printed and not returned.
 
----
+### All tests pass, but I get a timeout {#pass-but-timeout}
+
+## Sometimes users are confused when their solution is valid, passes all tests, test output panel is green, but they cannot submit the solution because tests time out. Such a situation means that solution does _not_ pass all the tests. Most likely, the solution returns a valid answer and passes tests up to some point, but it's too slow and then gets forcibly aborted by the runner due to exceeded execution time. Solution needs to be improved and made more performant. Some hints can be found [here](#timeout).
```
hobovsky, line 57:
```suggestion
Sometimes users are confused when their solution is valid, passes all tests, test output panel is green, but they cannot submit the solution because tests time out. Such a situation means that solution does _not_ pass all the tests. Most likely, the solution returns a valid answer and passes tests up to some point, but it's too slow and then gets forcibly aborted by the runner due to exceeded execution time. Solution needs to be improved and made more performant. Some hints can be found [here](#timeout).
```

PR #298, opened 2021-04-20
Add search with DocSearch

### LINKED ISSUE #4: Use DocSearch
Use [DocSearch](https://docsearch.algolia.com/) when we have contents written. The current search works only for the headers from `MarkdownPage`.

kazk:
Unfortunately, we didn't qualify for the free hosted version, but we can [run the crawler](https://docsearch.algolia.com/docs/run-your-own) every day in GitHub Actions. Hopefully, the free plan is enough for us.

Closes #4

### DIFF
--- a/.github/workflows/docsearch.yml
+++ b/.github/workflows/docsearch.yml
@@ -1,0 +1,22 @@
+name: DocSearch Scraper
+
+on:
+  schedule:
+    # Run daily at 00:00 UTC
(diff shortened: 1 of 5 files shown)


kazk:
For some reason, the search crashes when there's a recent search. To reproduce, first search something, and go to the page. Then open the search again and see the previous search under "Recents". Typing something crashes the page.

Issue #208, opened 2020-12-30
Create a guide for authoring in C++


hobovsky:
Create an article similar to https://docs.codewars.com/languages/python/authoring/ but for C++.

Points needing particular attention:
- includes (missing includes, C includes)
- random utilities
- stringizers
- custom assertion messages (modifications to the testing framework: https://github.com/codewars/snowhouse/pull/3 )
- compilation warnings
- input and output values: const ref vs value, replacements for arrays and `std::vector`
- avoid C: strings, arrays
- input modification, changing the signature of solution function

hobovsky:
@kazk could you share some insight on how solutions in C++ are built? What gets concatenated with what, what exactly is in the template (some includes? `main`?), etc?

error256:
From https://github.com/codewars/runner/issues/35#issuecomment-751105985
```cpp
#include <igloo/igloo_alt.h>
#include <igloo/CodewarsTestListener.h>

using namespace igloo;

// #include "preloaded.h" if preloaded

// [solution]

// [tests]

int main(int, const char *[]) {
  NullTestResultsOutput output;
  TestRunner runner(output);
  CodewarsTestListener listener;
  runner.AddListener(&listener);
  runner.Run();
}
```

hobovsky:
Would it still work if it looked like this:

```cpp

// #include "preloaded.h" if preloaded

// [solution]

#include <igloo/igloo_alt.h>
using namespace igloo;

// [tests]

#include <igloo/CodewarsTestListener.h>

int main(int, const char *[]) {
  NullTestResultsOutput output;
  TestRunner runner(output);
  CodewarsTestListener listener;
  runner.AddListener(&listener);
  runner.Run();
}
```

Not that I am proposing such change at the moment, I am just trying to figure out dependencies between snippet and what can potentially pollute what etc.

wtlgo:
@hobovsky asked me to share my thoughts about the proper usage of the `<random>` library.
So, here they are
- The C-way of generating random data `rand() % (a - b) + b` should be avoided unless it's really necessary. It's outdated.
- PRNG should be defined once per `Describe` unless, of course, it's necessary to have several PRNGs.
[...]

PR #311, opened 2021-04-27
C++ authoring tutorial

hobovsky:
Discussion: #208 
Deployment preview: [here](https://deploy-preview-311--reverent-edison-2864ea.netlify.app/languages/cpp/authoring/index)


content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,296 @@
...
+~~~if-not:cpp
+
+...text not visible in C++ description...
+
+~~~
+```
+
+## Tasks and Requirements
+
+Some concepts don't always translate well to or from C++. C++ allows for a variety of paradigms and techniques, and many of them are not widely spread accross other languages. Because of this, some constructs should be avoided and some translations just shouldn't be done.
```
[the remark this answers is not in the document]
hobovsky:
```suggestion
Some concepts don't always translate well to or from C++. C++ allows for a variety of paradigms and techniques, and many of them are not widely spread across other languages. Because of this, some constructs should be avoided and some translations just shouldn't be done.
```

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,369 @@
...
+- Many features of C++ do not have direct equivalents in other popular languages: mixture of allowed paradigms, template meta-programming, unmanaged memory backed by RAII, native access to the platform and runtime, and many others. C++ kata which rely on them can be difficult to translate to other languages.
+
+
+## Kata snippets and template
+
+Due to complicated compilation model of C++ code, paired with the fact that Codewars strictly enforces names of source files, their amount, and meaning, code runner uses a `[template]()` (TODO: document C++ runner template) to concatenate all kata snippets into a single translation unit. This fact has various consequences on the code, some positive, and some negative: namespace pollution, symbols introduced by one snippet being visible in other snippets, etc. However, in majority of cases, it does not affect kata in any way, and whenever possible, this behavior should be treated as implementation detail of C++ code runner. Kata snippets should be treated as separate source files, if possible.
```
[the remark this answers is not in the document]
hobovsky:
```suggestion
Due to complicated compilation model of C++ code, paired with the fact that Codewars strictly enforces names of source files, their amount, and meaning, code runner uses a `[template]()` (TODO: document C++ runner template) to concatenate all kata snippets into a single translation unit. This fact has various consequences on the code, some positive, and some negative: namespace pollution, symbols introduced by one snippet being visible in other snippets, etc. However, in the majority of cases, it does not affect kata in any way, and whenever possible, this behavior should be treated as implementation detail of C++ code runner. Kata snippets should be treated as separate source files, if possible.
```

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,296 @@
...
+## Coding
+
+### Code style
+
+C++ programmers have many sets of naming conventions or code style guides. Some of them can be found for example [here](https://isocpp.org/wiki/faq/coding-standards), or [here](https://google.github.io/styleguide/cppguide.html). Codewars does not strictly enforce any of them, just use whatever set of guidelines you like, but when you do, use it consistently. 
+
+
+### Header files
+
+C++ authors often forget to include required header files, or just leave them out deliberately because "it works" even when some of the files are not included. It happens mostly due to the following reasons:
```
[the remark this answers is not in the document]
hobovsky:
```suggestion
C++ authors often forget to include required header files, or just leave them out deliberately because "it works" even when some files are not included. It happens mostly due to the following reasons:
```

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,359 @@
...
+    Expected: equal to [ [unsupported type], [unsupported type] ]
+    Actual: [ [unsupported type], [unsupported type] ]
+```
+
+To rectify such issue in your tests, you can make such types suitable for stringification in a way described in dedicated documentation article on [custom Snowhouse stringizers](/languages/cpp/igloo/stringizers/).
+
+
+### Random utilities
+
+Until C++11, the most common way of producing random values was `rand` function. However, it has a set of problems: it needs to be properly seeded, and it's difficult to produce values outside of `0...RAND_MAX` range. Since C++11, standard library offers a set of functionalities which are designed to produce random values of different types, from various ranges, and with better distribution when compared to `rand`. Unfortunately, API presented in `random` header seems to be confusing and difficult to use and accompanied by amount of misconceptions, and authors either use it incorrectly, or resort to (not) good, old `rand`. However, working with `random` turns out to be not that difficult:
```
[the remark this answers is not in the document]
hobovsky:
```suggestion
Until C++11, the most common way of producing random values was `rand` function. However, it has a set of problems: it needs to be properly seeded, and it's difficult to produce values outside `0...RAND_MAX` range. Since C++11, standard library offers a set of functionalities which are designed to produce random values of different types, from various ranges, and with better distribution when compared to `rand`. Unfortunately, API presented in `random` header seems to be confusing and difficult to use and accompanied by amount of misconceptions, and authors either use it incorrectly, or resort to (not) good, old `rand`. However, working with `random` turns out to be not that difficult:
```

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,369 @@
...
+
+The reference solution or data ___must not___ be defined in the [Preloaded code](/authoring/guidelines/preloaded/).
+
+
+### Input mutation
+
+General guidelines for submission tests contain a section related to [input mutation](/authoring/guidelines/submission-tests/#input-mutation) and how to prevent users from abusing it to work around kata requirements. It's usually not a problem in "real world" C++ programming, but on Codewars, users can take advantage of vulnerable test suites and modify their behavior. Constness of a function argument can be forcefully cast away by a user, or they can change the function signature and remove `const` qualifier from input parameter(s).  After calling a user solution, tests should not rely on the state of such values and they should consider them as potentially modified by a user.
```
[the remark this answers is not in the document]
hobovsky:
```suggestion
General guidelines for submission tests contain a section related to [input mutation](/authoring/guidelines/submission-tests/#input-mutation) and how to prevent users from abusing it to work around kata requirements. It's usually not a problem in "real world" C++ programming, but on Codewars, users can take advantage of vulnerable test suites and modify their behavior. Constness of a function argument can be forcefully cast away by a user, or they can change the function signature and remove `const` qualifier from input parameter(s).  After calling a user solution, tests should not rely on the state of such values, and they should consider them as potentially modified by a user.
```

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,369 @@
...
+- Overloads of `Assert::That` which accept `message_supplier` should be preferred. Assertion message should provide meaningful details on the cause of failure, like test input, etc.
+- Some additional attention should be paid to the order of parameters passed to `Assert::That`. It differs between various assertion libraries, and it happens to be quite often confused by authors, mixing up `actual` and `expected` in assertion messages. For the C++ testing framework, the order is `(actual, SomeConstraint(expected))`.
+- To avoid unexpected crashes in tests, it's recommended to perform some additional assertions before assuming that the answer returned by the user solution has some particular form, or size. For example, if the solution returns a pointer, an explicit assertion should be added to check whether the returned pointer is valid, and not, for example, `nullptr`; the size of the returned container should be verified before accessing an element which could turn out to be located outside of its bounds.
```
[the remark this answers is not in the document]
hobovsky:
```suggestion
- To avoid unexpected crashes in tests, it's recommended to perform some additional assertions before assuming that the answer returned by the user solution has some particular form, or size. For example, if the solution returns a pointer, an explicit assertion should be added to check whether the returned pointer is valid, and not, for example, `nullptr`; the size of the returned container should be verified before accessing an element which could turn out to be located out of its bounds.
```

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,304 @@
...
+  double array[1000];     //small enough to be allocated on the stack,
+  double reference[1000]; //but you can use dynamic memory if necessary.
+  
+  for(int i=0; i<10; ++i) {
+    
+    //generate test cases
+    size_t array_size = rand() % 200 + 801;
+    fill_random_array(array, array_size);
+    
+    //since original array is no used after tests, it's enough to create only one copy
```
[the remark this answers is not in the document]
hobovsky:
```suggestion
    //since original array is not used after tests, it's enough to create only one copy
```

awesomeAD1 (changes requested):

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,359 @@
...
+~~~if-not:cpp
+
+...text not visible in C++ description...
+
+~~~
+```
+
+## Tasks and Requirements
+
+Some concepts don't always translate well to or from C++. C++ allows for a variety of paradigms and techniques, and many of them are not widely spread across other languages. Because of this, some constructs should be avoided and some translations just shouldn't be done.
```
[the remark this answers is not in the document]
hobovsky:
```suggestion
Some concepts don't always translate well to or from C++. C++ allows for a variety of paradigms and techniques, many of which are not necessarily suited to other languages. For this reason, certain constructs should be avoided and some kata just should not be translated to C++.
```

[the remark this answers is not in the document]
awesomeAD1: Perfect.

content/languages/cpp/authoring/index.md
```diff
@@ -93,15 +95,54 @@ Sometimes authors consider C++ just "C, but with classes". While C and C++ are c
...
 
 
-## Tests
+### `using` directives
 
-### Testing framework
+To avoid problems with namespace pollution, C++ snippets should not contain `using` directives (and especially `using namespace std;`) anywhere in the global namespace. Incorrect use of `using` can sometimes cause difficult to diagnose compilation problems. For example, `using namespace std;` placed incorrectly in "Complete solution" or in "Initial solution" snippet can result in failed compilation for _some_ users. To minimize such risks, authors should stick to following guidelines:
+- `using` directives should not be present in global scope of "Preloaded", "Complete solution", and "Initial solution" snippets. They are heavily discouraged in global scope of "Sample tests" and "Submission tests" snippets. This rule applies not only to `using namespace std;`, but also to other namespaces and namespace-qualified names.
+- `using` directives can be perfectly used in function scope: inside of an "Initial solution" function, in test helpers, or in `It` blocks.
```
awesomeAD1, line 105:
```suggestion
- `using` directives are perfectly fine inside the scope of an "Initial solution" function, test helpers, or `It` blocks.
```

Possibly even: "`using` directives are perfectly fine **when enclosed** inside the scope..."

content/languages/cpp/authoring/index.md
```diff
@@ -93,15 +95,54 @@ Sometimes authors consider C++ just "C, but with classes". While C and C++ are c
...
+
+Describe(TestGroup_2) {
+
+  It(TestCase_1) {
+    //...
+  }
+};
+```
+
+`Describe` blocks are expanded into C++ `struct`, and they can contain anything a structure can, in particular `public` and `private` access specifiers, member fields and functions, etc. `Describe` blocks can be also nested, with one remark: nested `Describe` sections compile and get registered for execution correctly, but test output panel does not report them hierarchically: all of them are flattened into one-level test report.
```
awesomeAD1, line 135:
```suggestion
`Describe` blocks are expanded into C++ `struct`s, and they can contain anything a `struct` can, in particular `public` and `private` access modifiers, member fields and functions, etc. `Describe` blocks can also be nested, with one caveat: while nested `Describe` sections are correctly compiled and registered for execution, the test output panel does not report them hierarchically; they are all flattened into a single-level test report.
```

awesomeAD1: Also: Consider "member fields and **methods**", if appropriate and correct.

content/languages/cpp/authoring/index.md
```diff
@@ -93,15 +95,54 @@ Sometimes authors consider C++ just "C, but with classes". While C and C++ are c
...
+
+### Snowhouse
+
+`Snowhouse` provides `Assert::That` function, which can accept either a constraint which has to be fulfilled by an asserted value (`Assert::That(actual, Equals(expected))`), or a fluent expression accepting actual value as input and returning a boolean value indicating a success or failure (`Assert::That(actual, Is().Equal(expected))`). Both types of assertions are equivalent to each other and authors can choose whichever suits them better. For clarity, this document uses only constraints as examples, but each of them has corresponding expression. Most useful ones are:
```
awesomeAD1, line 139:
```suggestion
`Snowhouse` provides the function `Assert::That`, which can accept either a constraint which has to be fulfilled by an asserted value (`Assert::That(actual, Equals(expected))`), or a fluent expression accepting actual value as input and returning a boolean value indicating success or failure (`Assert::That(actual, Is().Equal(expected))`). Both types of assertions are equivalent to each other and authors can choose whichever suits them better. For clarity, this document uses only constraints as examples, but each of them have corresponding expressions. The most useful ones are:
```

content/languages/cpp/authoring/index.md
```diff
@@ -172,9 +213,12 @@ for(int i=0; i<input_length; ++i) {
 //... testing logic    
 ```
 
+Standard library also contains some functions which can be helpful when generating random inputs:
```
awesomeAD1, line 216:
```suggestion
The standard library also contains some functions which can be helpful when generating random inputs:
```

content/languages/cpp/authoring/index.md
```diff
@@ -218,142 +261,118 @@ C++ sometimes requires some boilerplate code to implement non-trivial tests, che
...
-Test(fixed_tests, empty_array) {
-  
-  const double dummy = 42.5;
-  double items[1] = { dummy };
+    vector<int> items    = { 0, 1, 2, 3,  4 };  
+    vector<int> expected = { 0, 1, 4, 9, 16 };
+      
+    auto actual = square_every_item(items);
+    
+    //Assertion constraint checking for container equality
```
awesomeAD1, line 292:
```suggestion
    // Assertion constraint checking for container equality
```

content/languages/cpp/authoring/index.md
```diff
@@ -218,142 +261,118 @@ C++ sometimes requires some boilerplate code to implement non-trivial tests, che
...
-  double expected[5] = { 0.0, 1.21, 4.84, 10.89, 19.36 };
+  //a test case of fixed_tests suite for primary scenario
+  It(ExampleArray) {
     
-  square_every_item(items, 5);
-  
-  //assertion macro suitable for arrays of doubles
-  cr_assert_arr_eq_cmp(items, expected, 5, cmp_double_fuzzy_equal);
-}
+    //using directive limited to non-global scope
```
awesomeAD1, line 284:
```suggestion
    // using directive limited to non-global scope
```

content/languages/cpp/authoring/index.md
```diff
@@ -218,142 +261,118 @@ C++ sometimes requires some boilerplate code to implement non-trivial tests, che
...
-  srand(time(NULL));
-}
 
-//a test case of fixed_tests suite for primary scenario
-Test(fixed_tests, example_array) {
+Describe(FixedTests) {
   
-  double items[5]    = { 0.0, 1.1,  2.2,  3.3,   4.4 };  
-  double expected[5] = { 0.0, 1.21, 4.84, 10.89, 19.36 };
+  //a test case of fixed_tests suite for primary scenario
```
awesomeAD1, line 281:
```suggestion
  // a test case of fixed_tests suite for a primary scenario
```

content/languages/cpp/authoring/index.md
```diff
@@ -218,142 +261,118 @@ C++ sometimes requires some boilerplate code to implement non-trivial tests, che
...
+    auto actual = square_every_item(items);
+    
+    //Assertion constraint checking for container equality
+    Assert::That(actual, EqualsContainer(expected), ExtraMessage("Invalid answer for { 0, 1, 2, 3, 4 }"));
+  }
   
-  square_every_item(items, 0);
-  cr_assert_eq(items[0], dummy, "Empty array should not be tampered with.");
-}
+  //a test case of fixed_tests suite for potential edge case
```
awesomeAD1, line 296:
```suggestion
  // a test case of fixed_tests suite for a potential edge case
```

content/languages/cpp/authoring/index.md
```diff
@@ -218,142 +261,118 @@ C++ sometimes requires some boilerplate code to implement non-trivial tests, che
...
   
-  double input[10];
-  double actual[10];  
-  double reference[10];
+  std::mt19937 engine{ std::random_device{}() };
+  std::function<int   ()> gen_number     = std::bind(std::uniform_int_distribution<int   >{  1, 100 }, engine);
+  std::function<size_t()> gen_small_size = std::bind(std::uniform_int_distribution<size_t>{  2,  10 }, engine);
+  std::function<size_t()> gen_large_size = std::bind(std::uniform_int_distribution<size_t>{ 80, 100 }, engine);
+
+  //random test case generator
```
awesomeAD1, line 315:
```suggestion
  // random test case generator
```

content/languages/cpp/authoring/index.md
```diff
@@ -218,142 +261,118 @@ C++ sometimes requires some boilerplate code to implement non-trivial tests, che
...
-    
-    //generate test cases
-    size_t array_size = rand() % 200 + 801;
-    fill_random_array(array, array_size);
-    
-    //since original array is not used after tests, it's enough to create only one copy
-    memcpy(reference, array, sizeof(double) * array_size);
+public:  
+  
+  //a set of small random tests, with verbose debugging messages
```
awesomeAD1, line 335:
```suggestion
  // a set of small random tests, with verbose debugging messages
```

content/languages/cpp/authoring/index.md
```diff
@@ -218,142 +261,118 @@ C++ sometimes requires some boilerplate code to implement non-trivial tests, che
...
+  It(SmallArrays) {
     
-    square_every_item_ref(reference, array_size);
-    square_every_item(array, array_size);
+    for(int i=0; i<10; ++i) {
+      
+      //generate test case
```
awesomeAD1, line 340:
```suggestion
      // generate test case
```

content/languages/cpp/authoring/index.md
```diff
@@ -218,142 +261,118 @@ C++ sometimes requires some boilerplate code to implement non-trivial tests, che
...
+      
+      auto actual = square_every_item(input);
+  
+      //assertion uses custom message to avoid confusing test output
+      //it also uses data from original, non-mutated input array
+      Assert::That(actual, EqualsContainer(expected), ExtraMessage(fmt::format("Input: {}", stringify_input(original))));
+    }
+  }  
+
+  //a set of large random tests, with not so detailed debugging messages
```
awesomeAD1, line 357:
```suggestion
      // tests need to copy the input vector, because
      // it is used after calling the user and reference solution
      auto original = input;
      auto expected = square_every_item_ref(input);      
      
      auto actual = square_every_item(input);
  
      // assertion uses a custom message to avoid confusing test output.
      // it also uses data from original, non-mutated input array
      Assert::That(actual, EqualsContainer(expected), ExtraMessage(fmt::format("Input: {}", stringify_input(original))));
    }
  }  

  // a set of large random tests, with less detailed debugging messages
```

content/languages/cpp/authoring/index.md
```diff
@@ -218,142 +261,118 @@ C++ sometimes requires some boilerplate code to implement non-trivial tests, che
...
+  It(LargeArrays) {
     
-    //assertion uses custom message
-    cr_assert_arr_eq_cmp(array, reference, array_size, cmp_double_fuzzy_equal, "Invalid answer for arrays of size %zu", array_size);
+    for(int i=0; i<10; ++i) {
+      
+      //generate test cases
```
awesomeAD1, line 362:
```suggestion
      // generate test cases
```

content/languages/cpp/authoring/index.md
```diff
@@ -218,142 +261,118 @@ C++ sometimes requires some boilerplate code to implement non-trivial tests, che
...
+      size_t input_size = gen_large_size();
+      auto input = generate_random_input(input_size);
+      
+      auto expected = square_every_item_ref(input);
+      auto actual = square_every_item(input);
+      
+      //assertion uses custom message supplier
```
awesomeAD1, line 369:
```suggestion
      // assertion uses custom message supplier
```

awesomeAD1 (changes requested):
I made some suggestions in the example code sections as well: use a space after the double-backslashes in comments:
```
\\ good comment.
\\bad comment
```
This is preference, I'm sure, thus safely ignored.

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,378 @@
...
+
+
+## Kata snippets and template
+
+Due to the complicated compilation model of C++ code, paired with the fact that Codewars strictly enforces the names, amount, and meaning of source files, the code runner uses a `[template]()` (TODO: document C++ runner template) to concatenate all kata snippets into a single translation unit. This fact has various consequences on the code, some positive, and some negative: namespace pollution, symbols introduced by one snippet being visible in other snippets, etc. However, in the majority of cases, it does not affect kata in any way, and whenever possible, this behavior should be treated as an implementation detail of the C++ code runner. Kata snippets should be treated as separate source files, if possible.
+
+One consequence of using a template is that the signature of a solution function can be modified by the user in a way which can affect tests. To avoid the possibility of users tampering with the prototype of a solution function, it's recommended to re-declare it in the tests snippet.
```
error256, line 57: Redeclaration is questionable. It can be convenient to choose between accepting input by reference and by value (when a solution needs a copy anyway) and I think the only thing it prevents is lvalue-rvalue overloading that can be used to distinguish between fixed and random tests, which isn't really useful except when the tests are bad in the first place.

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,378 @@
...
+### Code style
+
+C++ programmers have many sets of naming conventions or code style guides. Some of them can be found for example [here](https://isocpp.org/wiki/faq/coding-standards), or [here](https://google.github.io/styleguide/cppguide.html). Codewars does not strictly enforce any of them, just use whatever set of guidelines you like, but when you do, use it consistently. 
+
+There are a few C++ coding guidelines which are violated by kata authors and translators particularly often:
+- Arguments which are not cheap to copy should be passed by const reference, and not by value.
+- Appropriate containers should be used, _especially_ when used as input arguments or return values:
+  - C-style raw arrays and pointers to arrays need to be avoided
+  - `std::array`, `std::pair`, or `std::tuple` should be used for data of known size. Avoid representing 2D points or pairs with `std::vector`.
+  - `std::vector` should be used for arrays of varying size
```
error256, line 73: I don't like calling non-arrays arrays, especially talking about languages that have arrays (although I see it in Python all the time anyway).
And I think the old normal C++ way would be a pair of iterators in many cases. Now that I think about it, it may actually be a good idea to update to C++20 now and use ranges and `span`s for input.

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,378 @@
...
+  - `std::string` should be used instead of C-strings
```
error256, line 74: And `std::string_view`. `std::string` may have some useful functions for some use cases, but I think `string_view` often is the best choice, I've been replacing `const std::string&` arguments in my solutions (at least in some of) to `std::string_view`s for a while now.
(OK, I saw `std::string_view` below...)

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,378 @@
...
+### Compilation warnings
+
+Compiler options related to warnings used by the C++ runner are somewhat strict and require some discipline to get the code to compile cleanly. `-Wall` and `-Wextra` may cause numerous warnings and some of them are very pedantic. However, code of C++ kata should still compile cleanly, without any warnings logged to the console. Even when a warning does not cause any problem with tests, users get distracted by them and blame them for failing tests.
+
+
+### Avoiding C
+
+Sometimes authors consider C++ just "C, but with classes". While C and C++ are compatible in many ways, that kind of perception is wrong: incorrect use of C features in C++ code leads to bad code at best, or to undefined behavior and difficult to diagnose errors in more extreme cases. Features and idioms from C language should be replaced with their equivalents from modern C++:
+
+- C features must not be used where they do not have well-defined behavior in C++. For example, memory management generally should not be done with `malloc`/`free` or similar, and VLAs should be replaced with `std::array` or `std::vector`.
```
error256, line 93: I don't think `std::array` can replace VLA, can it?

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,378 @@
...
+- C++ features should be preferred over those inherited from C, for example:
+  - C-style casts should be replaced with their C++ equivalents.
```
error256, line 95: This kind of thing looks to me like "we'll use this super-scalable DB that Google uses instead of a simple text file just because they say the former is cool and the latter isn't, although it's just 100 records used by 1 script".

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,378 @@
...
+//etc.
+
+//for convenience, distributions can be bound with PRNG
+auto gen_length = std::bind(rand_length, engine);
+auto gen_letter = std::bind(rand_letter, engine);
+
+std::string input;
+//input string generator: string of random length, composed of random letters
+size_t input_length = gen_length(); //use length generator
+for(int i=0; i<input_length; ++i) {
```
error256, line 208: `size_t i`? Comparing signed with unsigned will probably generate a warning. (I guess `int input_length` is OK too for kata-scale sizes if it appears in an existing translation, but for best practices it should be `size_t`.)

hobovsky: 🤦 
```suggestion
for(size_t i=0; i<input_length; ++i) {  
```

content/languages/cpp/authoring/index.md
hobovsky: You mean specifically casts, or some other points from this paragraph too?

content/languages/cpp/authoring/index.md
hobovsky: I was wondering what can be a replacement for VLAs and could not find any. But still, VLAs are not valid C++, are they?

content/languages/cpp/authoring/index.md
hobovsky:
```suggestion
  - `std::string` or `std::string_view` should be used instead of C-strings
```

content/languages/cpp/authoring/index.md
hobovsky: You are not a first person to raise this issue, but as much as I can see where you guys are coming from, I do not like it ;)

My main idea behind this point was to prevent situations when a user changes const ref to non-const ref and mutates inputs to modify behavior of test cases. Or when user changes a narrow type to a wider type to work around some potential difficulties of solution.

On a side not related to kata, I do not like the idea of changing the interface just for a sake of an implementation detail. I know this technique and I know that it's used, I am just not sure whether a public interface of a solution is a good place to apply it.

I will think how to reword this part. Maybe something along the lines that _"make sure that users won't be able to break tests by modifying a signature of solution function"_, or something like that?

content/languages/cpp/authoring/index.md
hobovsky:
```suggestion
  - `std::vector` is not the only available option, and can be replaced by another container (for example, `std::set`), a set of iterators, a range, or any other construct suitable for the task.
```

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,378 @@
...
+### Compilation warnings
+
+Compiler options related to warnings used by the C++ runner are somewhat strict and require some discipline to get the code to compile cleanly. `-Wall` and `-Wextra` may cause numerous warnings and some of them are very pedantic. However, code of C++ kata should still compile cleanly, without any warnings logged to the console. Even when a warning does not cause any problem with tests, users get distracted by them and blame them for failing tests.
+
+
+### Avoiding C
+
+Sometimes authors consider C++ just "C, but with classes". While C and C++ are compatible in many ways, that kind of perception is wrong: incorrect use of C features in C++ code leads to bad code at best, or to undefined behavior and difficult to diagnose errors in more extreme cases. Features and idioms from C language should be replaced with their equivalents from modern C++:
+
+- C features must not be used where they do not have well-defined behavior in C++. For example, memory management generally should not be done with `malloc`/`free` or similar, and VLAs should be replaced with `std::array` or `std::vector`.
```
hobovsky, line 93:
```suggestion
- C features must not be used where they do not have well-defined behavior in C++. For example, memory management generally should not be done with `malloc`/`free` or similar, and VLAs should be replaced with `std::vector`.
```

content/languages/cpp/authoring/index.md
hobovsky:
```suggestion
One consequence of using a template is that the signature of a solution function can be modified by the user in a way which can affect tests. Make sure that tests are not possible to work around by users tampering with the prototype of the solution function. One possible way to ensure this is to re-declare the solution function in the tests snippet.
```

Do you think it's better now?

content/languages/cpp/authoring/index.md
hobovsky: I removed this part. If you think something should be added back and/or reworded, let me know.

content/languages/cpp/authoring/index.md
```diff
@@ -52,11 +50,11 @@ Some concepts don't always translate well to or from C++. C++ allows for a varie
...
 
-One consequence of using a template is that the signature of a solution function can be modified by the user in a way which can affect tests. To avoid the possibility of users tampering with the prototype of a solution function, it's recommended to re-declare it in the tests snippet.
+One consequence of using a template is that the signature of a solution function can be modified by the user in a way which can affect tests. Make sure that tests are not possible to work around by users tampering with the prototype of the solution function. One possible way to ensure this is to re-declare the solution function in the tests snippet.
```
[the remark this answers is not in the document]
hobovsky: I removed the mentions of re-declaration of user solution.

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,387 @@
+---
+title: Authoring C++ Content
+sidebar_label: Authoring
```
kazk, line 3:
```suggestion
sidebar_label: Authoring
slug: /languages/cpp/authoring
```

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,387 @@
...
+- Many features of C++ do not have direct equivalents in other popular languages: mixture of allowed paradigms, template meta-programming, unmanaged memory backed by RAII, native access to the platform and runtime, and many others. C++ kata which rely on them can be difficult to translate to other languages.
+
+
+## Kata snippets and template
+
+Due to the complicated compilation model of C++ code, paired with the fact that Codewars strictly enforces the names, amount, and meaning of source files, the code runner uses a [template](/languages/cpp/solution_template) to concatenate all kata snippets into a single translation unit. This fact has various consequences on the code, some positive, and some negative: namespace pollution, symbols introduced by one snippet being visible in other snippets, etc. However, in the majority of cases, it does not affect kata in any way, and whenever possible, this behavior should be treated as an implementation detail of the C++ code runner. Kata snippets should be treated as separate source files, if possible.
```
kazk, line 54: I don't know why it was decided to concatenate at first, but I don't think there's any reason to do this (other than we have 1K+ kata that relies on the setup). Maybe I'm missing something.

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,388 @@
...
+One consequence of using a template is that the signature of a solution function can be modified by the user in a way which can affect tests. Make sure that tests are not possible to work around by users tampering with the prototype of the solution function.
+
+Another issue caused by the use of the template is excessive namespace pollution. See paagraph on [`using` directives`](#using-directives) for more details.
+
+
+## Coding
+
+### Code style
+
+C++ programmers have many sets of naming conventions or code style guides. Some of them can be found for example [here](https://isocpp.org/wiki/faq/coding-standards), or [here](https://google.github.io/styleguide/cppguide.html). Codewars does not strictly enforce any of them, just use whatever set of guidelines you like, but when you do, use it consistently.
```
kazk, line 65: Try avoiding `[here]` links.

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,388 @@
...
+Another issue caused by the use of the template is excessive namespace pollution. See paagraph on [`using` directives`](#using-directives) for more details.
+
+
+## Coding
+
+### Code style
+
+C++ programmers have many sets of naming conventions or code style guides. Some of them can be found for example [here](https://isocpp.org/wiki/faq/coding-standards), or [here](https://google.github.io/styleguide/cppguide.html). Codewars does not strictly enforce any of them, just use whatever set of guidelines you like, but when you do, use it consistently. 
+
+There are a few C++ coding guidelines which are violated by kata authors and translators particularly often:
```
kazk, line 67: Doesn't this contradict with the above?

> Codewars does not strictly enforce any of them, just use whatever set of guidelines you like, but when you do, use it consistently.

Maybe we should have a small set of rules that content creators are expected to follow based on this list.

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,387 @@
...
+...text not visible in C++ description...
+
+~~~
+```
+
+## Tasks and Requirements
+
+Some concepts don't always translate well to or from C++. C++ allows for a variety of paradigms and techniques, many of which are not necessarily suited to other languages. For this reason, certain constructs should be avoided and some kata just should not be translated to C++.
+- C++ is statically typed, so any task that depends on dynamic typing can be difficult to translate into C++, and attempts to force a C++ kata to reflect a dynamically typed interface can lead to ideas that enforce a really bad design.
+- There are only a few additional libraries available to the C++ runner, so kata that take advantage of specialized packages installed for other languages become much more difficult to implement in C++.
```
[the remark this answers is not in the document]
hobovsky: Still,  not as much as very rich libraries of languages like Java, .Net. Framework, Python with things like big numbers, crypto (hash functions), parsers and serializers (XML, csv), web connectivity, etc. There are kata related to all these elements, but they are not many, so maybe indeed it's not worth mentioning.

sidebars.js
```diff
@@ -164,9 +164,11 @@ module.exports = {
           label: "C++",
           items: [
             "languages/cpp/index",
+            "languages/cpp/authoring/index",
+            "languages/cpp/solution_template",
             {
               type: "category",
-              label: "igloo",
+              label: "Testing Framework",
```
kazk, line 171: Let's do the following instead

1. Use label "Igloo" ("Testing Framework" sounds awkward)
2. Move `languages/cpp/igloo` to `languages/cpp/igloo/index` with `slug: /languages/cpp/igloo` and `sidebar_label: Overview`.

content/languages/cpp/authoring/index.md
hobovsky: I meant the first paragraph to be related to things like naming conventions, placement of braces, and other minuscule things people like to fight about over the Internet. 
The latter paragraph i meant to relate to more serious / not just aestethic language best practices.
Maybe my wording is off, and just do not know how to express myself clearly.

content/languages/cpp/solution_template.md
```diff
@@ -0,0 +1,32 @@
+---
```
kazk, line 1: Use hyphen in file names.

content/languages/cpp/authoring/index.md
hobovsky: I can see a few reasons for this, but I believe the most probable one is handling of templates in C++. Templates have to be defined in every translation unit, and that's why they're always located in header files. If a solution would need to define a template, it would be not enough for a solution to be a cpp file (i.e. a separate translation unit).

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+To rectify this issue in your tests, you can make such types suitable for stringification in a way described in the dedicated documentation article on [custom Snowhouse stringizers](/languages/cpp/igloo/stringizers/).
+
+
+### Random utilities
+
+Until C++11, the most common way of producing random values was the `rand` function. However, it has a set of problems: it needs to be properly seeded, and it is difficult to produce values outside of `0...RAND_MAX` range. Since C++11, the standard library offers a set of functionalities which are designed to produce random values of different types, from various ranges, and with better distribution when compared to `rand`. Unfortunately, the API presented in the `random` header seems to be confusing and difficult to use and accompanied by a few misconceptions, and authors either use it incorrectly or resort to (not) good, old `rand`. However, working with `random` turns out to be not that difficult:
+
+```cpp
+//use random_device only as a seeder
```
kazk, line 185: Can we add a space after `//`? It's difficult to read.

I also think it's more common to have a space in any language.

```javascript
// comment
//comment
``` 
```python
# comment Python mentions this in PEP8 too.
#comment
```
```haskell
-- comment
--comment
```
```ocaml
(* comment *)
(*comment*)
```

kazk: I think it's difficult for me to read because my brain tries to ignore it. I've only seen no space when commenting out the line.

For the following code:
```javascript
//foo
bar();
```
I read:
```javascript
bar();
```
until I try not to ignore.

That doesn't happen when it has a space:
```javascript
// foo
bar();
```

kazk: Found https://stackoverflow.com/questions/1467058/space-between-line-comment-characters-and-start-of-actual-comment and the answers matches my experience.

I'll make it a rule. In our docs, space is required.

hobovsky: Done.

content/languages/cpp/solution_template.md
hobovsky: Done.

sidebars.js
hobovsky: Done.

content/languages/cpp/authoring/index.md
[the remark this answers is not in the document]
hobovsky: I removed this point.

content/languages/cpp/authoring/index.md
hobovsky: Changed.

content/languages/cpp/authoring/index.md
hobovsky: I tried to reword both points, I am not sure if it's better though.

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,387 @@
...
+
+
+### Input mutation
+
+General guidelines for submission tests contain a section related to [input mutation](/authoring/guidelines/submission-tests/#input-mutation) and how to prevent users from abusing it to work around kata requirements. It's usually not a problem in "real world" C++ programming, but on Codewars, users can take advantage of vulnerable test suites and modify their behavior. Constness of a function argument can be forcefully cast away by a user, or they can change the function signature and remove the `const` qualifier from input parameter(s).  After calling a user solution, tests should not rely on the state of such values, and they should consider them as potentially modified by the user.
+
+
+### Calling assertions
+
+Snowhouse provides a set of useful [assertions](/languages/cpp/igloo#snowhouse-assertions)`, but when used incorrectly, they can cause a series of problems:
```
error256, line 235:
```suggestion
Snowhouse provides a set of useful [assertions](/languages/cpp/igloo#snowhouse-assertions), but when used incorrectly, they can cause a series of problems:
```

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,387 @@
...
+- Use of an assertion not suitable for the given case may lead to incorrect test results,
+- Incorrectly used assertions may produce confusing or unhelpful messages.
+
+To avoid the above problems, calls to assertion functions should respect the following rules:
+- The expected value should be calculated _before_ invoking an assertion. The `expected` parameter passed to the assertion should not be a function call expression, but a value calculated directly beforehand. _(? does this point hold for C++ kata?)_
```
error256, line 240: > does this point hold for C++ kata

If it's about hiding the name of the reference solution, then it depends on whether it can be called from users' solutions. If it's about input mutation, it depends on whether the input is copied for the reference solution. If everything is done the best way, it shouldn't matter.

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,387 @@
...
+- Appropriate assertion functions should be used for each given test. `AssertEquals` is not the only option, and Snowhouse provides a selection of constraints and expressions suitable for many scenarios: `EqualsWithDelta` for floating point comparisons, `EqualsContainer ` to compare containers with a predicate, etc.
```
error256, line 241:
```suggestion
- Appropriate assertion functions should be used for each given test. `AssertEquals` is not the only option, and Snowhouse provides a selection of constraints and expressions suitable for many scenarios: `EqualsWithDelta` for floating point comparisons, `EqualsContainer` to compare containers with a predicate, etc.
```

error256 (changes requested):

content/languages/cpp/authoring/index.md
hobovsky:
```suggestion
```

I think will remove it for now, and we will see if reviewers will be able to handle issues related to both above aspects correctly. Or maybe just make this point a recommendation instead of `should`?

error256 (approved):

content/languages/cpp/authoring/index.md
kazk: Ah, I see. If it was possible to have `solution.h` and `solution.cpp`, the test can include `solution.h` to be compiled separately and linked. Since we can't do that, we don't have a choice.

content/languages/cpp/solution-template.md
```diff
@@ -0,0 +1,32 @@
...
+---
+
+# C++ Solution Setup
+
+_TODO: add details_
+
+
+## Solution template
+
+```cpp
```
kazk, line 13: To make it easier to test locally, the solution template can be changed to the following that can be reused:
```cpp
// main.cpp
#include <igloo/igloo_alt.h>
#include <igloo/CodewarsTestListener.h>
using namespace igloo;

// Use `#include` to "copy" contents here
#include "preloaded.h" // use empty file if there's no preloaded
#include "solution.cpp"
#include "tests.cpp"

int main(int, const char *[]) {
  NullTestResultsOutput output;
  TestRunner runner(output);
  CodewarsTestListener listener;
  runner.AddListener(&listener);
  runner.Run();
}
```

content/languages/cpp/authoring/index.md
kazk: I don't think we need a separate page for the template snippet. There's nothing much to explain further, and the new reusable one is self-explanatory.

hobovsky: Solution snippet is not the only one affected by this issue. It applies also to, for example, preloaded, which might want to define global/static data in a cpp part, and declare it in a header file.

> Ah, I see. If it was possible to have `solution.h` and `solution.cpp`, the test can include `solution.h` to be compiled separately and linked. Since we can't do that, we don't have a choice.

It _could_ be done, but in kinda awkward way: snippets could contain some kind of markers which would separate "header" part and ".cpp part", and runner would split them into files appropriately. It's clunky, awkward, but in 99% not necessary, and some default behavior would suffice (no markers and store everything as a cpp file, compile, and link). Only those who need it, could use something like:

```cpp
/*============ solution.h ============*/
// ... declarations, template definitions

/*===========solution.cpp============*/
// ... implementation part
```

Not that I like this solution, but it _is_ one of solutions.



> I don't think we need a separate page for the template snippet. There's nothing much to explain further, and the new reusable one is self-explanatory.

Ok so I will remove the link to the "solution template" page for now, and later I will turn the "solution template" page into "how to train on cpp kata locally" and add the link back.

content/languages/cpp/authoring/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+
+## Kata snippets and template
+
+Due to the complicated compilation model of C++ code, paired with the fact that Codewars strictly enforces the names, amount, and meaning of source files, the code runner uses a template to concatenate all kata snippets into a single translation unit. This fact has various consequences on the code, some positive, and some negative: namespace pollution, symbols introduced by one snippet being visible in other snippets, etc. However, in the majority of cases, it does not affect kata in any way, and whenever possible, this behavior should be treated as an implementation detail of the C++ code runner. Kata snippets should be treated as separate source files, if possible.
+
+One consequence of using a template is that the signature of a solution function can be modified by the user in a way which can affect tests. Make sure that tests are not possible to work around by users tampering with the prototype of the solution function.
+
+Another issue caused by the use of the template is excessive namespace pollution. See paagraph on [`using` directives`](#using-directives) for more details.
```
kazk, line 57:
```suggestion
Another issue caused by the use of the template is excessive namespace pollution. See paragraph on [`using` directives`](#using-directives) for more details.
```

kazk (approved):

PR #319, opened 2021-05-08
Update ranks.mdx

julmoser:
table reference was missing, added reference to score table as I didn't find one matching the exact reference


Blind4Basics:
meerged, thx.

PR #324, opened 2021-05-18
add info about assertion message for RSpec

Blind4Basics:


content/languages/ruby/index.md
```diff
@@ -18,9 +18,9 @@ From Ruby 3.0, [RSpec](https://rspec.info/) is used under the hood:
...
 describe "Example" do
   it "should return the sum" do
-    expect(add(1, 1)).to eq(2)
+    expect(add(1, 1)).to eq(2), "failure message"
     # The following is still supported, but new tests should not use them.
-    # Test.assert_equals(add(1, 1), 2)
+    # Test.assert_equals(add(1, 1), 2, "failure message")
   end
 end
 ```
```
Blind4Basics, line 26:
````suggestion
```
Note that if an assertion message is provided while using RSpec, actual and expected results have to be in that message otherwise the information will not be shown to the user.
````

PR #333, opened 2021-06-12
add info controlling size of svg 

Blind4Basics:


content/authoring/recipes/adding-images.md
```diff
@@ -78,6 +78,7 @@ Codewars allows a safe subset of HTML in Markdown including `<svg>`, so it's pos
 
 The image markup can be pasted directly into the kata editor without additional processing or encoding, and usually it's more readable than encoded image data.
 But still, even simple-looking SVGs often require a significant amount of code, inflating the kata description. However, they can be minified or optimized with some additional tools (for example [SVGOMG][svgomg]), or by removing unnecessary parts of markup and Metadata. Additionally, the results of mixing HTML with Markdown are ill-defined and subject to change.
+Note that those images are scaling automatically in the description page, which can become very annoying. To avoid this, the `svg` code can be wrapped with  a `<div style="width:400px;height:400px">...</div>` tag, choosing the dimension that suites the situation better.
```
kazk, line 81: Why a wrapper? Can't you just add `style` to `svg`?

Blind4Basics: errrr... I just didn't think about that XD
Editing

Blind4Basics, line 81:
```suggestion
Note that those images are scaling automatically in the description page, which can become very annoying. To avoid this, the `style` attribute of the `svg` tag can be added/updated  with `style="width:...px;height:...px"` , choosing the dimension that suites the situation better.
```

Blind4Basics: see below, plz

kazk, line 81:
```suggestion
Make sure to prevent images from becoming too large by using `width/height` attributes, `style` attributes, or a wrapper element with `style`.
```

This applies to any image, so it's probably better to mention this as general advice before listing the options (around line 23).

PR #367, opened 2022-01-08
Added documentation for Factor

Kacarott:
I roughly used the existing JavaScript guide as a template, and also added some Factor specific content.

I thought perhaps nomennescio could look over it for errors or changes, but it might be easier to "pull now, fix later" so that he can read directly in the docs.

Preview: [here](https://deploy-preview-367--reverent-edison-2864ea.netlify.app/languages/factor/authoring)

### DIFF
--- a/content/languages/factor/authoring.md
+++ b/content/languages/factor/authoring.md
@@ -83,44 +83,46 @@ To create and present test output, testest uses the `describe#{` and `it#{` word
 ```factor
 "Fixed tests" describe#{
   "Odd numbers" it#{
-    ! some assertions...
+    ! some tests...
   }#
 
   "Even numbers" it#{
-    ! some assertions...
+    ! some tests...
   }#
 }#
 
 "Random tests" describe#{
   "Small inputs" it#{
-    ! some assertions...
+    ! some tests...
   }#
 
   "Large inputs" it#{
-    ! some assertions...
+    ! some tests...
   }#
 }#
 ```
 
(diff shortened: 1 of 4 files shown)


Blind4Basics:
the "review dog" left some warning in the "files changed" page. Could you handle them? ([see there](https://github.com/codewars/docs/pull/367/files))

Do you know how to update the PR?

error256:
Should the new page be added to `sidebars.js` too or how does this work?

Kacarott:
Ah yeah it should be, probably? I assumed it was automatic based on the directory structure and files within.

I can update this later this afternoon.

kazk:
> I had assumed it was automatic based on the directory structure and files within.

It can be automatic in the newer version of Docusaurus. We should update...

nomennescio:
@Kacarott 

Overall, your Factor descriptions are agreeable, but I have one clear point for improvement.

For the `testest` framework you describe (unit)tests as "assertions", whereas they are actual test(case)s with two possible outcomes : success or failure. It's also important to describe that each testcase `<{ inputs -> expected-outputs }>` is actually an evaluation of ALL `inputs` AND ALL `expected-outputs`, which can contain words (functions) that executes, and which passes if the resulting stack is the same, and fails otherwise. For reasons of reporting of failing test cases, it might be beneficial to have additional arguments on the stack. These will be passed to the `error` object in fields `got` and `expected` and can be parsed in the custom error reporting quotation in var `test-failed.` In case of success, the quotation in var `test-passed.` is called, with no arguments.

It might help to add some examples of such usage. I can dig up some of that on request.

PR #397, opened 2022-06-13
Add Python unittest

ggorlen:


ggorlen:
Is it worth encouraging unittest now instead of the CW testing framework (and/or deprecating the latter)?

content/languages/python/unittest.mdx
```diff
@@ -0,0 +1,64 @@
+---
+title: Python unittest
+sidebar_label: unittest
+kind: reference
+---
+
+import ThemedImage from "@theme/ThemedImage";
+
+Codewars offers Python's [unittest](https://docs.python.org/3/library/unittest.html) module for testing solutions. Codewars uses a custom reporter for unittest to format the output for the code runner. Visit the [python-test-framework](https://github.com/codewars/python-unittest) GitHub repository for more information.
```
kazk, line 9: We don't _support_ `unittest` to test kata solutions. It's only supported in Kumite. It happens to work if you add `unittest.main(testRunner=CodewarsTestRunner())` because the test is run with `python tests.py`. The runner still does stuff like prepending `import codewars_test as test` because it's asked to use the Codewars test framework.

I really want to make test framework configurable per kata and stop using the custom test framework though.

JohanWiltink:
> Is it worth encouraging unittest now instead of the CW testing framework (and/or deprecating the latter)?

After people fixed thousands upon thousands Python kata for the last framework, modules, and Python 3.10, I'm not sure if deprecating any of that will go over well.

content/languages/python/unittest.mdx
ggorlen: Ah, maybe I got excited too soon! Can I use this to test, for example, Pandas challenges, or is it recommended to stick with CW test framework?

ggorlen:
Also, I notice many testing frameworks are missing guides. I can add these pretty easily from Qualified's docs, but maybe I should defer to someone more familiar with the platform?

content/languages/python/unittest.mdx
kazk: I just remembered we discussed this a few years ago. https://github.com/codewars/python-test-framework/pull/7#issuecomment-647050696

I don't want to recommend using `unittest` now, but if you need it, I guess it's fine to use it with a comment at the top explaining the hack.

kazk:
> but maybe I should defer to someone more familiar with the platform?

No, we can review them and suggest changes if it's awkward in Codewars context. PRs welcomed.

content/languages/python/unittest.mdx
ggorlen: Yep, I saw your recent push in exo and figured it was an option with 3.10.

kazk:
@ggorlen Please open separate PRs, so it's easier to review and merge individually.

PR #399, opened 2022-06-18
Flesh out PHPUnit

ggorlen:


kazk (approved):

content/languages/php/phpunit.md
```diff
@@ -9,6 +9,28 @@ tags:
...
+    return $a + $b;
+}
+```
+
+```php
+class AddTest extends TestCase {
+    public function testAdd() {
+        $expected = 3;
+        $actual = add(1, 2);
+        $this->assertEquals($expected, $actual);
```
error256, line 29: Tests like this with non-strict equality can be passed by `return true;` for all non-zero values.

ggorlen: Sorry, I don't think I understand. Isn't `assertEquals` strict equality?

error256: No, strict equality is `assertSame`.

ggorlen: I'm still not following the consequence of this. Can you provide an example where `assertEquals` is problematic to use for CW?

[`assertSame`](https://phpunit.readthedocs.io/en/9.5/assertions.html#assertsame) seems to be identity of objects, not what I would want:

> Reports an error identified by `$message` if the two variables `$expected` and `$actual` do not reference the same object.

error256: It's problemattc pretty much with every function returning numbers:
https://www.codewars.com/kata/reviews/57b89247f07f80714400001e/groups/62adf0ca916c01000120c085

ggorlen: OK, thanks--that's pretty clearly broken. Great catch. So it seems that `assertEquals` is best for object equalty and `assertSame` for everything else, including arrays?

error256: It seems so, but I don't really know PHP, so I can't tell much.

ggorlen: Neither do I, clearly. I'll add a PR for that, reference this conversation and hopefully someone can validate it if there's any issues. Thanks again--super helpful.

Issue #375, opened 2022-02-07
Recipe: parametrized tests in JUnit 5


hobovsky:
```java
import java.util.*;
import java.util.stream.*;
import java.util.function.*;

import org.junit.jupiter.api.*;
import org.junit.jupiter.params.*;
import org.junit.jupiter.params.provider.*;
import static org.junit.jupiter.api.Assertions.*;

// Values source
@ParameterizedTest
@ValueSource(strings = {"", "  "})
void isBlank_ShouldReturnTrueForNullOrBlankStrings(String input) {
    assertTrue(Strings.isBlank(input));
}

// CSV source
@ParameterizedTest
@CsvSource({"test,TEST", "tEst,TEST", "Java,JAVA"})
void toUpperCase_ShouldGenerateTheExpectedUppercaseValue(String input, String expected) {
    String actualValue = input.toUpperCase();
    assertEquals(expected, actualValue);
}

// Generator method
@ParameterizedTest
@MethodSource("provideStringsForIsBlank")
void isBlank_ShouldReturnTrueForNullOrBlankStrings(String input, boolean expected) {
    assertEquals(expected, Strings.isBlank(input));
}

private static Stream<Arguments> provideStringsForIsBlank() {
    return Stream.of(
      Arguments.of(null, true),
      Arguments.of("", true),
      Arguments.of("  ", true),
      Arguments.of("not blank", false)
    );
}
```

Madjosz:
I would suggest to add also display names:

```java
import org.junit.jupiter.api.DisplayName;

@DisplayName("Tests with JUnit 5")
class SampleTests {

    @ParameterizedTest(name = "input: '{0}' and I expect you to return >>{1}<<")
    @DisplayName("We test some fixed Strings")
    @CsvSource({"test,TEST", "tEst,TEST"})
    void toUpperCase_ShouldGenerateTheExpectedUppercaseValue(String input, String expected) {
        String actualValue = input.toUpperCase();
        assertEquals(expected, actualValue);
    }
}
```

This will lead to the following output:
![image](https://user-images.githubusercontent.com/28844868/152800035-22c11028-d9ef-4dca-af2d-ef0e7083469e.png)

Madjosz:
```java
import org.junit.jupiter.api.MethodOrderer;
import org.junit.jupiter.api.Order;
import org.junit.jupiter.api.TestMethodOrder;

@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
class SampleTests {
    @Test
[...]

PR #400, opened 2022-06-18
Flesh out JUnit

ggorlen:


ggorlen:
Parameterized tests and method ordering from https://github.com/codewars/docs/issues/375 would be useful to add.

kazk (approved):

content/languages/java/junit.md
```diff
@@ -2,3 +2,61 @@
 title: JUnit
 kind: reference
 ---
+
+You can use JUnit [4](https://junit.org/junit4/) or [5](https://junit.org/junit5/) in Codewars. 5 offers nesting, pretty name display and multiple test classes.
+
+Note that JUnit assertions use `(expected, actual)` parameter ordering rather than the typical `(actual, expected)`.
```
hobovsky, line 8: This statement is biased :) After years of work with JUnit and NUnit I was surprised that anything uses `(actual, expected)`

content/languages/java/junit.md
```diff
@@ -2,3 +2,61 @@
 title: JUnit
 kind: reference
 ---
+
+You can use JUnit [4](https://junit.org/junit4/) or [5](https://junit.org/junit5/) in Codewars. 5 offers nesting, pretty name display and multiple test classes.
```
hobovsky, line 6: Technically, multiple test classes are possible with JUnit 4 on Codewars too. It is clunky because of how test snippet is handled, but I remember I managed to use multiple test classes by nesting them inside of a top-level class.

content/languages/java/junit.md
hobovsky: OK now I double checked and I see that NUnit uses `(actual, expected)` too.
It's actually interesting how I manage to switch between JUnit and NUnit and get the order of arguments right, but I can never remember it for Chai or Python framework.

ggorlen: Yeah, it's a bit biased. I think JUnit and PHPUnit are the "only" (I'm sure there are others that I'm not familiar with) that use (exp, act), off the top of my head. We can still make the statement more neutral, but I tend to see more issues with PHP having asserts in the wrong direction than other languages--again, totally anecdotal. Feel free to PR a fix.

content/languages/java/junit.md
ggorlen: I'd be happy to see that. Feel free to PR!

hobovsky: BTW do we even need to mention JUnit 4 in Codewars docs? It cannot be removed because there's many kata using it, but why not deprecate it and provide docs only for JUnit 5, in a similar way Javascript Codewars framework was available but discouraged and not extensively documented?

ggorlen: I thought about that but wasn't sure I wasn't overreaching. If that's the general sentiment in the community (you're more in it than I), I'm all for stripping out the JUnit 4 stuff.

PR #401, opened 2022-06-19
Avoid assertEquals and clarify PHP

ggorlen:
See comments in #399: https://github.com/codewars/docs/pull/399#pullrequestreview-1011406669. This PR clarifies the surprising behavior of `assertEquals` and provides a few alternatives.

### DIFF
--- a/content/languages/php/phpunit.md
+++ b/content/languages/php/phpunit.md
@@ -9,6 +9,28 @@ tags:
 
 # PHPUnit
 
+All [PHPUnit](https://phpunit.readthedocs.io/) tests start with a subclass of `TestCase`. You can then add one or more test case methods to that class, each of which must be public and start with `test`. In Codewars' PHP versions 7.4+, PHPUnit requires the name of the test class to end with `Test`.
+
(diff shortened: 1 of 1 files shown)


kazk (approved):
:exploding_head: I need to update the examples and template used in Codewars...

content/languages/php/phpunit.md
```diff
@@ -26,11 +24,60 @@ class AddTest extends TestCase {
...
+### Float equality
+
+For float comparisons, [`assertEqualsWithDelta`](https://phpunit.readthedocs.io/en/9.5/assertions.html#assertequalswithdelta) has similar unpredictable behavior as `assertEquals`:
+
+```php
+$this->assertEqualsWithDelta(0.99, "1", 0.1); // passes
+$this->assertEqualsWithDelta(99999, true, 0.1); // passes
+```
+
+Implementing a custom delta assertion based on `assertTrue` may be a safer bet.
```
error256, line 78: A custom assertion would only differ in checking the type. But then why not just check the type first and then use `assertEqualsWithDelta`?

error256 (approved):

PR #403, opened 2022-06-20
Flesh out Pester example

ggorlen:


content/languages/powershell/pester.md
```diff
@@ -9,6 +9,36 @@ tags:
...
+```
+
+### Test Fixture
+
+```powershell
+BeforeAll {
+    . $PSCommandPath.Replace(".Tests.ps1", ".ps1")
+    # Or
+    # . $PSScriptRoot/Solution.ps1
+}
```
kazk, line 29:
```suggestion
BeforeAll {
    . $PSCommandPath.Replace(".Tests.ps1", ".ps1")
}
```
I think it's better to stick to this one.

kazk (approved):

content/languages/powershell/pester.md
```diff
@@ -9,6 +9,34 @@ tags:
...
+### Test Fixture
+
+```powershell
+BeforeAll {
+    . $PSCommandPath.Replace(".Tests.ps1", ".ps1")
+}
+
+Describe "Add-Numbers" {
+    It "adds positive numbers" {
+        Add-Numbers 1 1 | Should -Be 2
```
hobovsky, line 31: Does the framework has a way to specify custom assertion messages?

ggorlen: It seems so: https://pester.dev/docs/assertions/#because

hobovsky: Okay, I'll ask in another way: 
I am not sure whats exactly the scope of these "Flesh out X" PRs, but do you think it would be better to describe assertions in a way richer than absolute working minimum, which effectively results in suboptimal feedback of tests?
I try to actively discourage laziness and promote writing tests in a way so they give as useful feedback as possible, and i think that showing how to achieve this right from the beginning would hopefully result in a better quality of tests created in the future.

kazk: Yeah, it's a good idea to show a minimal example of how it should be used on Codewars. The page can also include more details as needed.

ggorlen: Thanks for the feedback.

I find that the default assertions provide optimal messages for most testing libraries and use cases, so adding custom assertions might actually harm output readability of a simple integer, string or array comparison.

I'd love to write comprehensive guides for each testing library, but given limited time it seems best to provide a minimal working example of as many as I can offer, and allow others to flesh them out as they see fit (particularly if they're domain experts), or following an expanded standard format of some kind. Most of the docs are adapted from https://docs.qualified.io/ so it's pretty quick for me.

Most testing libraries have detailed documentation for special assertions and uncommon case use cases, so it seems unnecessary to reiterate all of that in CW docs, especially since it can go out of date quickly. The primary goal for these PRs is (as you mention) to show a minimal example that wires things together for CW, which may involve special imports or other boilerplate code that isn't available from the library's docs.

In other cases, what the framework docs recommend isn't necessarily apprpriate for CW, or is a common mistake we see, like incorrect order of `expected, actual` parameters. Those seem worth documenting here as well.

In any case, the current state of affairs is blank pages for almost all testing libraries which seemed like it needed action.

How does all of this sound to you?

hobovsky: First of all, don't get me wrong: I am not trying to fight for anything particular right now in this thread, I just hoped I found a way for a potentially large gain (tests with useful feedback) bought with a relatively low effort (one additional parameter in examples of assertions). 

Remainder of the post is not trying to prove any point, but attempts to show what kind of challenges Codewars kata face.

> I find that the default assertions provide optimal messages for most testing libraries and use cases, so adding custom assertion messages might actually harm test case and output readability of a simple integer, string or array comparison.

Objection. First of all, there are frameworks notorious for their useless assertion messages. C kata are full of Criterion's `expression ((strcpm((kataSolution(arg1, arg2)), referenceSolution(arg1, arg2))) == 0) is false`, Swift has `left not equal to right`, no one in OCaml knows how to use `~printer` and `~msg` so by default assertion message is literally `" not equal "`, literally no one I asked knows how to provide messages in Haskell, etc. Additionally, even when default assertion message is quite clear `Expected X, but got Y`, it's very often missing any kind of context, because authors do not provide any context in test output. After encountering an error I know that Y is incorrect and X is expected, but... why? Custom assertion messages are one of ways to provide a context of a failure.

> I'd love to write comprehensive guides for each testing library, but given limited time it seems best to provide a minimal working example of as many as I can offer, and allow others to flesh them out as they see fit (particularly if they're domain experts), or following an expanded standard format of some kind (which doesn't exist yet as far as I can tell). Most of the docs are adapted from https://docs.qualified.io/ so it's pretty quick for me.

There's a few "Authoring kata in language X" articles, which focus also on tools useful for tests: various assertions, generated test cases, whatever testing framework provides and can be useful. The articles are often large and it;s perfectly fine to get to them iteratively with time. I dont mean in any way that we should create full articles right from the beginning.

> Most testing libraries have detailed documentation for special assertions and uncommon case use cases, so it seems unnecessary to reiterate all of that in CW docs, especially since it can go out of date quickly. The primary goal for these PRs is (as you mention) to show a minimal example that wires things together for CW, which may involve special imports or other boilerplate code that isn't available from the library's docs.

Yes and no. It's been already a couple of times when a translator answered that _"I did not know the `cr_assert_str_eq` exists"_, and such. Additionally, what I hoped would be possible by remarks from previous post, is to get a very useful thing by a small effort. Because adding a single parameter to an example _is_ a small effort, isn't it? Yeah, I am totally asking from a position of a person not knowing what exactly is necessary to get this, sorry.

> In other cases, what the library docs recommend isn't necessarily apprpriate for CW, or is a common mistake we see, like incorrect order of `expected, actual` parameters. Those CW-specific best practices seem worth documenting here as well.

This is exactly what I mean.
Codewars is a kind of a special beast, because translations are created by two types of people: a) those who know the target language, testing framework, and its intricacies, and b) [overenthusiastic] learners, be it experienced coders or not, who translate _because they can_, because they hope to bring more attention to their kata, or _hey, points!_. The former type is not a problem, as such people usually know what they are doing. The latter, on the other hand, can cause various kind of issues, usually due to picking the route which is the easiest to follow, or the easiest to find. We can create such a route by feeding them with docs which not only explain what and how, but also, to some extent, promote good Codewars (and real life, maybe) practices. Let's take away from them the weapon of _"I followed docs so it's good enough"_. People tend to stick to the first practice they encounter, and when the first thing they see is the minimal working example, they will not accept any critique of it later. When they learn from a richer example, they will use it and pass it further.

Again, I understand that writing full blown manuals is not a goal here. But maybe it would be possible to have such a tiny tiny thing like example of custom messages in asserions.

ggorlen: No doubt that some assertion libraries have terrible default messages that must be overridden. Feel free to add them if you feel that it's helpful. But if at all possible, I prefer to rely on the default. This makes a consistent experience and reduces the visual noise in the test code. I've seen plenty of "creative" output assertions (e.g. colored text) that are worse UX than the default. Too much code/customization in test suites seems to be as serious a problem as not enough.

There are so many ways that careless users can screw things up that it'd take the mentioned "Authoring kata in language X" to begin to suss out all the potential pitfalls--if they read it. I'm not convinced that offering no guidance is better than a small example, though, so I don't forsee a negative impact on quality here.

PR #406, opened 2022-06-21
Add R basic setup

ggorlen:


hobovsky:
I will be more than happy to provide any feedback, just let me know when you'd want me to stop :)
Please also mind that I am not familiar with this many languages as some other users, and that I live in another timezone so it can take some time until I get to review a freshly published PR.
As for other users, you can use language rank leaderboards to get idea who's proficient with what language. Unnamed is always a safe bet, this guy was born and raised in a programming equivalent of the Tower of Babel.

kazk:
Leaderboard for R: https://www.codewars.com/api/v1/leaders/ranks/r

(I haven't added UI because it's inaccurate due to a known issue preventing contributors from getting rank scores after contributing, but it's usable.)

hobovsky:
Maybe it would be useful to have a link to the framework docs (https://testthat.r-lib.org/) so users could explore further if necessary? And a version number of the testing framework as well?
Or you think it's not necessary, or an additional maintenance problem?

ggorlen:
> Maybe it would be useful to have a link to the framework docs (https://testthat.r-lib.org/) so users could explore further if necessary? And a version number of the testing framework as well? Or you think it's not necessary, or an additional maintenance problem?

I'm flexible--the link is a nice gesture and might save users a search, but if we do decide to include it, we'd want to be consistent and apply to all testing libraries. It's a bit more maintenance, and I'm not sure the benefit is huge.

kazk:
We can also move that section to this page. I guess that's better.

ggorlen:
> We can also move that section to this page. I guess that's better.

I think so, so it's consistent with languages that have multiple test libraries.

kazk (approved):

PR #413, opened 2022-07-09
Before start

hobovsky:
Page which attempts to explain some intricacies of Codewars to absolute beginners and users not really familiar with coding.

Preview: [here](https://deploy-preview-413--reverent-edison-2864ea.netlify.app/training/beginners/)
Preview of "training example": [here](https://deploy-preview-413--reverent-edison-2864ea.netlify.app/training/training-example/)


kazk:
I like the idea behind this page, but I think we need to present it better.

We should simply explain the differences from traditional coding challenge service that uses stdin/stdout by mentioning kata are unit tests, and how that affects you to write your solution.

hobovsky:
> We should simply explain the differences from traditional coding challenge service that uses stdin/stdout [...]

I don't like the idea of comparing Codewars to "traditional coding challenge service that uses stdin/stdout" because I think that users, especially the ones affected by the problem I am trying to address with this page, are not necessarily familiar with "traditional coding challenge service that uses stdin/stdout". It _is_ possible that some users are confused because they try to apply analogies from other OJs, but I think that for majority of newbies, printing and reading is the first thing they learned somewhere (with some introductory materials, on some bootcamps, courses, internet, whatever), and it's the only known way of interaction their code can use.

> [...] by mentioning kata are unit tests, and how that affects you to write your solution.

This is what I want to do, but initially I placed the page in another section of docs where mentioning tests did not really seem fit. Later I moved the page and here it's good place to mention tests, but I did not add it yet. I will think of something.

content/training/beginners.md
```diff
@@ -0,0 +1,32 @@
...
+
+Solution is writen by a user who attempts a kata. To work correctly, it has to conform to expectations of tests: it has to be a function if tests expect it to be one, or a class, or a variable. Statement of a kata task should clearly explain what is required. To make things easier for users, initial solution setup usually (but not always) contains a stub which presents an expected form of a solution which conforms to requirements imposed by tests. It is ready to be called, usually should not be changed by users, and just needs to be filled in. Additional hints about how a user solution should look like and how it is invoked can be found in example tests.
```
hobovsky, line 13:
```suggestion
Solution is written by a user who attempts a kata. To work correctly, it has to conform to expectations of tests: it has to be a function if tests expect it to be one, or a class, or a variable. Statement of a kata task should clearly explain what is required. To make things easier for users, initial solution setup usually (but not always) contains a stub which presents an expected form of a solution which conforms to requirements imposed by tests. It is ready to be called, usually should not be changed by users, and just needs to be filled in. Additional hints about how a user solution should look like and how it is invoked can be found in example tests.
```

content/training/beginners.md
```diff
@@ -0,0 +1,32 @@
...
+ - **How functions return results they came up with?** When your function is run, it's expected by tests to return an answer to its caller, so it could be checked and tested. In particular, printing the answer to a console, or anywhere else, will not work as expected.
+- **How are functions invoked?** - tests will call your function, possibly multiple times. Every time, they will pass in different input arguments, and will inspect an answer returned by it. There is no need for users to call their function themselves, and there's no need for any kind of entry point (like some `main` function, or equivalent) just to make a solution runnable. Tests already know how to run solutions. Users also need to make sure that their solution can be run more than once, and every time returns a correct answer.
+
+## Solution as a class
+
+Another popular kind of a solution is a class, or some function associated with a class (like a class method, or a prototype extention, etc.) In majority of cases, it's enough to fill in a solution template in a correct place to create a working solution. But sometimes, some additional knowledge might come in handy:
```
hobovsky, line 27:
```suggestion
Another popular kind of a solution is a class, or some function associated with a class (like a class method, or a prototype extension, etc.) In the majority of cases, it's enough to fill in a solution template in a correct place to create a working solution. But sometimes, some additional knowledge might come in handy:
```

content/training/beginners.md
```diff
@@ -0,0 +1,32 @@
...
+ - **How are classes constructed and how to create a new one?** - as much as it's helpful to know this in general, this part is usually handled by kata in a solution setup provided to users. In majority of cases a stub of a required classes is already provided and there should be no need to create one from scratch.
```
hobovsky, line 28:
```suggestion
 - **How are classes constructed and how to create a new one?** - as much as it's helpful to know this in general, this part is usually handled by kata in a solution setup provided to users. In the majority of cases a stub of a required class is already provided and there should be no need to create one from scratch.
```

hobovsky:
Since I got feedback from users that the page is generally useful, and questions from beginners pop up more and more often, I decided to merge it.  
I will fix problems as they appear.

PR #416, opened 2022-07-14
Move Test::More to its own file

ggorlen:


kazk:
I just noticed you've been adding `Test Fixture` header. Can we use something else? Maybe simply `Preloaded`, `Solution` and `Tests`?
Test fixture sounds awkward to me here, since it usually means to set up a known/fixed data. https://en.wikipedia.org/wiki/Test_fixture#Software

ggorlen:
> I just noticed you've been adding `Test Fixture` header. Can we use something else? Maybe simply `Preloaded`, `Solution` and `Tests`? Test fixture sounds awkward to me here, since it usually means to set up a known/fixed data. https://en.wikipedia.org/wiki/Test_fixture#Software

I'm not a fan of it either but it seemed to be the way things were elsewhere in the docs, so I didn't want to be inconsistent. I'll do another PR and update across the board to Tests.

hobovsky (commented):
I am sorry but I am not familiar enough to Perl to give any constructive feedback here :(

kazk:
> I'm not a fan of it either but it seemed to be the way things were elsewhere in the docs, so I didn't want to be inconsistent. I'll do another PR and update across the board to Tests.

Thanks. Yeah, we should make it consistent. I only found `fixture` in https://github.com/codewars/docs/blob/master/content/languages/c/authoring/index.md?plain=1#L94 ("test fixture setup") and that's probably used correctly.

kazk (approved):

ggorlen:
> > I'm not a fan of it either but it seemed to be the way things were elsewhere in the docs, so I didn't want to be inconsistent. I'll do another PR and update across the board to Tests.
> 
> Thanks. Yeah, we should make it consistent. I only found `fixture` in https://github.com/codewars/docs/blob/master/content/languages/c/authoring/index.md?plain=1#L94 ("test fixture setup") and that's probably used correctly.

Yep, I didn't see much of it in CW, so I must have brought it in from Qualified docs, where it's all over the place. I'll remove it there, too.

PR #417, opened 2022-07-14
Add basic Clojure test example

ggorlen:


content/languages/clojure/clojure-test.md
```diff
@@ -2,3 +2,27 @@
...
+(ns clojure.adder)
+
+(defn add [a b]
+  (+ a b))
+```
+
+### Tests
+
+```clojure
+(ns clojure.adder-test
```
kazk, line 22: We shouldn't use `clojure.` namespace for user's code.

ggorlen: I have no idea about Clojure beyond copy-paste from existing Qualified docs. What namespace do you suggest? Something to the effect of `adder`/`challenge`/`solution`?

kazk: Anything as long as it doesn't conflict with or belong to the standard library (`clojure`) and package namespaces.

I'd use `solution` and `solution-test` or `adder` and `adder-test`.

kazk (approved):

PR #419, opened 2022-07-14
Add basic Igloo example

ggorlen:


hobovsky:
Preview: [here](https://deploy-preview-419--reverent-edison-2864ea.netlify.app/languages/cpp/igloo)

hobovsky (approved):

ggorlen:
> I noticed the Perl page you added has `reference` tag, and this one on Igloo does not. Should it?

I haven't been touching the tags at all, but it's a good idea to make them consistent. I'll go through and do that, thanks!

PR #433, opened 2022-07-18
Add Groovy examples

ggorlen:


content/languages/groovy/spock.md
```diff
@@ -7,6 +7,30 @@ tags:
...
+```
+
+### Tests
+
+```groovy
+import spock.lang.Specification
+
+class AdderSpec extends Specification {
+  def "Adder.add returns the sum"() {
+    expect:
```
error256, line 27: What is this label for?

ggorlen: Spock has that in their docs: https://spockframework.org/spock/docs/2.1/data_driven_testing.html. I guess the idea is to separate different parts of the test case. Would you recommend removing it?

ggorlen: I'll resolve since the linked docs also have `expect:` labels pointing to basic assertions, but if a Kotlin expert arrives happy to have them PR a better version down the line. I'm mostly trying to get something minimal that's reasonably in the ballpark for each lang.

PR #444, opened 2022-09-02
Add basic Mocha example

ggorlen:
[Preview](https://deploy-preview-444--reverent-edison-2864ea.netlify.app/languages/javascript/mocha/)

### DIFF
--- a/content/languages/javascript/mocha.md
+++ b/content/languages/javascript/mocha.md
@@ -21,7 +21,7 @@ const {assert, config, expect} = require("chai");
 // don't truncate assertion diff output
 config.truncateThreshold = 0;
 
-// if applicable to the challenge, don't truncate printing deeply objects
+// if applicable to the challenge, don't truncate printing deeply-nested objects
 require("util").inspect.defaultOptions.depth = null;
 
 describe("adder", () => {
--- a/content/languages/javascript/mocha.md
+++ b/content/languages/javascript/mocha.md
@@ -24,7 +24,7 @@ config.truncateThreshold = 0;
 // if applicable to the challenge, don't truncate printing deeply objects
 require("util").inspect.defaultOptions.depth = null;
 
-describe("adder",() => {
+describe("adder", () => {
   it("should add two numbers", () => {
     assert.equal(add(1, 2), 3);
 
--- a/content/languages/javascript/mocha.md
+++ b/content/languages/javascript/mocha.md
@@ -21,9 +21,6 @@ const {assert, config, expect} = require("chai");
 // don't truncate assertion diff output
 config.truncateThreshold = 0;
 
-// if applicable to the challenge, don't truncate printing deeply-nested objects
-require("util").inspect.defaultOptions.depth = null;
-
 describe("add", () => {
   it("should add two numbers", () => {
     assert.strictEqual(add(1, 2), 3);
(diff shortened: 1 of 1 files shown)


content/languages/javascript/mocha.md
```diff
@@ -3,6 +3,36 @@ title: Mocha
...
+
+// don't truncate assertion diff output
+config.truncateThreshold = 0;
+
+// if applicable to the challenge, don't truncate printing deeply-nested objects
+require("util").inspect.defaultOptions.depth = null;
+
+describe("add", () => {
+  it("should add two numbers", () => {
+    assert.equal(add(1, 2), 3);
```
kazk, line 29: `assert.strictEqual` should be used.

content/languages/javascript/mocha.md
```diff
@@ -3,6 +3,36 @@ title: Mocha
...
+### Tests
+
+```javascript
+const {assert, config, expect} = require("chai");
+
+// don't truncate assertion diff output
+config.truncateThreshold = 0;
+
+// if applicable to the challenge, don't truncate printing deeply-nested objects
+require("util").inspect.defaultOptions.depth = null;
```
kazk, line 25: Not sure if we should recommend adding this in tests. Does this affect the failure message?

Blind4Basics: the `truncateThreshold` part? it's a must have. The vast majority of the kata are undebuggable without it.

ggorlen: I think kaz is referring to `require("util").inspect.defaultOptions.depth = null;`. I've been using it at Qualified for awhile and haven't seen any issues, but that doesn't necessarily mean it isn't causing a problem somewhere.

The problem it solves is that logging deeply-nested objects is truncated:

```js
Welcome to Node.js v18.4.0.
Type ".help" for more information.
> console.log({a:{a:{a:{a:{a:{}}}}}})
{ a: { a: { a: [Object] } } }
undefined
> require("util").inspect.defaultOptions.depth = null;
null
> console.log({a:{a:{a:{a:{a:{}}}}}})
{
  a: {
    a: { a: { a: { a: {} } } }
  }
}
```

But this doesn't seem to affect Chai assertion failure messages, which truncate the depth either way. So it's mainly a convenience I add to challenges with deeply-nested object structures for users who might `console.log()` to debug. I can leave it out if that's the consensus--they can use `console.dir(foo, {depth: null})` or add it themselves.

content/languages/javascript/mocha.md
ggorlen: You're right, and I'll have to change that at Q as well. Big oversight!

content/languages/javascript/mocha.md
Blind4Basics: Ah, I didn't remember how the comments are added in between/around lines of code, sorry...

ggorlen: No prob. Since it doesn't seem to affect the tests, I'll leave the suggestion out.

content/languages/javascript/mocha.md
```diff
@@ -3,6 +3,33 @@ title: Mocha
...
+
+// don't truncate assertion diff output
+config.truncateThreshold = 0;
+
+describe("add", () => {
+  it("should add two numbers", () => {
+    assert.strictEqual(add(1, 2), 3);
+
+    // ...or use expect assertions:
+    expect(add(1, 2)).to.eq(3);
```
kazk, line 29:
```suggestion
    expect(add(1, 2)).to.equal(3);
```

PR #457, opened 2022-10-18
Update index.md

jojo20071:
Added -math to Packages


hobovsky:
Why would `math` be even needed to be added to docs? Isn't it available always? Is it possible to not have `math`?

Blind4Basics:
yes it's a builting. Always there.

@jojo20071 : only packages that do not always come with the default installation are listed.

Closing & cheers

PR #472, opened 2023-02-18
Recipe: OUnit assertions

hobovsky:
Since authors often do not bother to check output of failing tests and are not aware of how bad the default assertion messages are with OUnit, I created this recipe to post it as links in problematic translations.

### DIFF
--- a/content/languages/ocaml/recipes.md
+++ b/content/languages/ocaml/recipes.md
@@ -1,0 +1,36 @@
+---
+title: OCaml recipes for Codewars
+tags: [authoring, recipe, ocaml]
+---
+
+
+## Informative assertion messages
+
+Default assertion messages of OCaml OUnit are usually very bad. OUnit assertons accept two optional, named arguments:
+- `~printer`, used to stringify values presented by assertion messages.
+- `~msg`, used to provide additional information about failure, if necessary.
+
+With both arguments used, test output becomes more explicit:
+
+```ocaml
+"(square 3) should return 9" >:: (fun _ ->
+  let actual = square 3 in
+  assert_equal 9 actual ~printer: string_of_int ~msg: "Incorrect answer for n=3"
+)
+```
+
+```text
+Time: 1083ms Passed: 0Failed: 1Exit Code: 1
+  Test Results:
+    Tests for square
+      (square 3) should return 9
+        Incorrect answer for n=3
+        expected: 9 but got: 10
+    Completed in 0.24ms
+  Completed in 0.26ms
+```
+
+### References
+
+- [Improving OUnit Output](https://cs3110.github.io/textbook/chapters/data/ounit.html#improving-ounit-output)
+- [OUnit: xUnit testing framework for OCaml](https://gildor478.github.io/ounit/ounit2/index.html#error-reporting)
--- a/sidebars.js
+++ b/sidebars.js
@@ -512,7 +512,10 @@ module.exports = {
             type: "doc",
             id: "languages/ocaml/index",
           },
-          items: ["languages/ocaml/ounit"],
+          items: [
+            "languages/ocaml/ounit",
+            "languages/ocaml/recipes"
+          ],
         },
         {
           type: "category",


content/languages/ocaml/recipes.md
```diff
@@ -0,0 +1,36 @@
...
+
+```ocaml
+"(square 3) should return 9" >:: (fun _ ->
+  let actual = square 3 in
+  assert_equal 9 actual ~printer: string_of_int ~msg: "Incorrect answer for n=3"
+)
+```
+
+```text
+Time: 1083ms Passed: 0Failed: 1Exit Code: 1
```
monadius, line 23: Do we need this line? If you decide to keep it, then please insert spaces before `Failed` and `Exit Code`.

hobovsky: Initially I wanted to add a screenshot because it would show quite well how bad the default messages are. But then it turned out that having screenshots in recipes would probably inflate amount of files significantly, especially when having two of them (dark mode and bright mode).  That's why I went with copy/paste of contents of the output panel.

Do you think it's OK this way? Or maybe unnecessary? Or maybe screenshots would be better?

I can remove the initial line, it's not exactly necessary.

monadius: I prefer plain text for the output (easier to maintain and modify if necessary). The output looks better without the initial line.

content/languages/ocaml/recipes.md
```diff
@@ -0,0 +1,34 @@
...
+        Incorrect answer for n=3
+        expected: 9 but got: 10
+    Completed in 0.22ms
+Completed in 0.24ms
+```
+
+### References
+
+- [Improving OUnit Output](https://cs3110.github.io/textbook/chapters/data/ounit.html#improving-ounit-output)
+- [OUnit: xUnit testing framework for OCaml](https://gildor478.github.io/ounit/ounit2/index.html#error-reporting)
```
monadius, line 34: I prefer the following link here: https://ocaml.org/p/ounit2/2.2.3/doc/index.html#error-reporting (looks more official).

content/languages/ocaml/recipes.md
```diff
@@ -0,0 +1,34 @@
+---
+title: OCaml recipes for Codewars
+tags: [authoring, recipe, ocaml]
+---
+
+
+## Informative assertion messages
+
+Default assertion messages of OCaml OUnit are usually very bad. OUnit assertons accept two optional, named arguments:
+- `~printer`, used to stringify values presented by assertion messages.
```
monadius, line 10: I recommend the following definition:
`~printer` defines a printer (stringifier) for compared values.

monadius, line 10: It would be useful to provide some common `printer` values: for printing `int`, `float`, `'a list`, `'a array`, etc. Printers can be defined with standard OCaml functions or with Batteries (or Core) functions. 

Since not all translators use Batteries, it may be preferable to use standard OCaml functions in this recipe. On the other hand, Batteries include many useful printing functions. Here is a function for printing `int list` written by trashy_incel:
```ocaml
let string_of_int_list (lst : int list) : string =
  "[" ^ String.concat "; " (List.map string_of_int lst) ^ "]"
```
And here is the same function using Batteries:
```ocaml
let string_of_int_list = IO.to_string (List.print Int.print)
```
It is very short and can be inlined.

What do you think? Should we include examples of common printers here? If yes, then is it better to use standard OCaml functions or functions from Batteries (I don't know Core so I cannot say if it is better or not)?

content/languages/ocaml/recipes.md
```diff
@@ -0,0 +1,34 @@
+---
+title: OCaml recipes for Codewars
+tags: [authoring, recipe, ocaml]
+---
+
+
+## Informative assertion messages
+
+Default assertion messages of OCaml OUnit are usually very bad. OUnit assertons accept two optional, named arguments:
```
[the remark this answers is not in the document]
hobovsky:
```suggestion
Default assertion messages of OCaml OUnit are usually very bad. OUnit assertions accept two optional, named arguments:
```

content/languages/ocaml/recipes.md
hobovsky: For me, your wording sounds more like "stringified values are compared" rather than "stringified values are presented in a message".

But I trust your experience more than my feelings, so I go with your proposal.

hobovsky: Initially I planned to present use of a custom printer, but decided not to (for now).

I like the idea of listing some commonly used printers, but I have one problem with this: I struggle with finding some balance between helping authors, and repeating "obvious" things which can be found in obvious places by anyone who would be bothered to search.

Nothing of this short recipe is a thing which, in my opinion, should be presented this explicitly. The recipe presents almost nothing specific to Codewars authoring, and everything here is practical aspects of the ecosystem. Heck, even if something is totally not familiar with OCaml, or OUnit, all it takes is one run of failing tests, a tiny piece of reflection that _"heck, this is bad, there has to be some way to make this look better"_, and a search in _official docs_ . I know why I have written this recipe, but why it has to be written is beyond me.

If authors struggle to find out that `~printer` even exists, then I think it's not reasonable to expect them to know how to create custom printers. I'd give them commonly used printers too. If you have an idea what printers, and ideas for implementation, just drop a list here, or on Discord, or in a ticket. I for example did not know what printer to use to print strings, or I would not know what to use to print a generic `'a list`. Whether to use Core or Batteries, I don't know because I am totally not familiar with OCaml ecosystem and I do not know what are common practices. For me, both are fine. I will rely on judgment of you and other users.

Thanks for reviewing!

content/languages/ocaml/recipes.md
```diff
@@ -31,4 +31,4 @@ Completed in 0.24ms
 ### References
 
 - [Improving OUnit Output](https://cs3110.github.io/textbook/chapters/data/ounit.html#improving-ounit-output)
-- [OUnit: xUnit testing framework for OCaml](https://gildor478.github.io/ounit/ounit2/index.html#error-reporting)
\ No newline at end of file
+- [OUnit: xUnit testing framework for OCaml](https://ocaml.org/p/ounit2/2.2.3/doc/index.html#error-reporting)
```
interrupt-routine, line 34: It seems that only the  `2.2.6` version of OUnit has docs on ocaml.org, all the other versions return a 404.

Therefore the version number should be changed in the URL: `2.2.3` --> `2.2.6`.
Same problem at the page  `https://docs.codewars.com/languages/ocaml`: it points to `2.2.7`

PR #486, opened 2023-11-26
Update index.md

Blind4Basics:
Fix Node versions

### DIFF
--- a/content/languages/javascript/index.md
+++ b/content/languages/javascript/index.md
@@ -11,7 +11,7 @@ tags: [javascript]
 - Node 8.x
 - Node 10.x
 - Node 12.x
-- Node 14.x
+- Node 18.x
 
 ## Test Frameworks
 


Blind4Basics:
@kazk I just realized that the page I'm trying to modifiy might be automatically generated from somewhere else... I guess that's the problem here?

error256:
Maybe check the log?
> 6:29:41 PM: [ERROR] Error: Invalid sidebar file at "sidebars.js".
> 6:29:41 PM: These sidebar document ids do not exist:
> 6:29:41 PM: - community/rules

Blind4Basics:
yeah, I saw that, but I didn't touch this, so no idea what I should be doing with that info.

kazk:
Looks like the build is failing since https://github.com/codewars/docs/commit/2b4ab11b2b73a1bfb91dec533626c2f76ee164eb because the `sidebars.js` is now referencing a document id that doesn't exist.

@VanessaGreene Please open a PR instead of pushing directly to master to avoid breaking. I think GitHub provides the option when editing the page in your browser. If that's not possible for some reason, please make sure to confirm the changes you made were applied, and let me know when it's not.

kazk:
@Blind4Basics Rebasing should fix the build. You can also create a new PR from the current master.

PR #492, opened 2023-12-27
Update RISC-V ISA documentation

StarsDown64:
Simple edit to reflect the extra extensions supported in codewars/riscv#7. Also adds links to specifications since some extensions are not included in the RISC-V Spec.
Still not completely happy with line 13, I feel like it's not visually appealing but the ISA string the closest there is to a version number. Maybe listing the QEMU version? Not sure.
As I will comment in codewars/riscv#7 as well, Zbkx is not necessary as it's in both Zk and Zks shorthands.

### DIFF
--- a/content/languages/riscv/index.md
+++ b/content/languages/riscv/index.md
@@ -29,10 +29,10 @@ Codewars uses [QEMU user mode emulation](https://www.qemu.org/docs/master/user/m
     - `Zks` - ShangMi
 
 The most recent version of the specification for each can be downloaded below (as of 26 Dec 2023):
-- <sup>*</sup><a href="https://github.com/riscv/riscv-isa-manual/releases/download/riscv-isa-release-056b6ff-2023-10-02/unpriv-isa-asciidoc.pdf" download rel="external" type="application/pdf">RISC-V Unprivileged Specification</a>
-- <sup>†</sup><a href="https://github.com/riscv/riscv-v-spec/releases/download/v1.0/riscv-v-spec-1.0.pdf" download rel="external" type="application/pdf">Vector Extension Specification</a>
-- <sup>‡</sup><a href="https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0.pdf" download rel="external" type="application/pdf">Bit-Manipulation Extension Specification</a>
-- <sup>§</sup><a href="https://github.com/riscv/riscv-crypto/releases/download/v1.0.1-scalar/riscv-crypto-spec-scalar-v1.0.1.pdf" download rel="external" type="application/pdf">Cryptography Extension Specification</a>
+- <sup>*</sup><a href="https://github.com/riscv/riscv-isa-manual/releases/download/riscv-isa-release-056b6ff-2023-10-02/unpriv-isa-asciidoc.pdf">RISC-V Unprivileged Specification</a>
+- <sup>†</sup><a href="https://github.com/riscv/riscv-v-spec/releases/download/v1.0/riscv-v-spec-1.0.pdf">Vector Extension Specification</a>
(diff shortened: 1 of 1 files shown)


StarsDown64:
@dramforever request your thoughts.

dramforever:
LGTM. I think for version string maybe QEMU as you said and also GCC and Binutils version?

StarsDown64:
Now that I think about it, none of the QEMU, GCC, or any runner tool version is specific to riscv language, so I wouldn't think it appropriate for the riscv specific docs.

dramforever:
My understanding is that GCC, Binutils, QEMU *are* in fact unique to `riscv` language. Presently at least, no other language uses QEMU for emulation, and the GCC and Binutils here are cross compilation toolchains, which also no other language uses.

StarsDown64:
Huh. I see. If you know where to look, could you get me any relevant version info to include? I'm not familiar at all with the runner or toolchain.

dramforever:
Just had an idea to "cheat" and do a `system("gcc --version"); system("as --version");` in a kumite and got 

```
gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
GNU assembler (GNU Binutils for Ubuntu) 2.38
```

As for QEMU I think this means we're using 7.1 https://github.com/codewars/riscv/pull/9

DonaldKellett (approved):
Very detailed, LGTM!

hobovsky:
Looking at the preview here: https://deploy-preview-492--reverent-edison-2864ea.netlify.app/languages/riscv I am not sure I like the style of the asterisk/dagger/etc. annotations. Formatting is a bit off, and I htink they might look somewhat unusual to people more familiar with just numbered annotations like <sup>1)</sup> . What do you think?

DonaldKellett:
The style seems in line with official RISC-V documentation, but I agree their appearance isn't that common in the Codewars docs and might not fit in well with the rest of the documentation.

I'm personally not bothered by it but let's collect some opinions from a few others before we decide whether to stick to the current style or revert to plain numbered annotations.

StarsDown64:
Is there precedent elsewhere in the docs? I searched through the repo and found one other instance of the single dagger (in content\gamification\priveleges.md:27,32) and none of the double dagger or section symbol. I tried looking for `\d\)` or `\[\d\]` but couldn't find any being used as footnote markers.

hobovsky:
Numbers vs daggers is one thing, and we can fix it later if needed. But another thing which puts me off a bit is that the footnote markers seem to be not superscripted,  neither in the main text, nor in footnotes themselves?

StarsDown64:
I see what you mean. I like the look of superscripts as well. Adding.

DonaldKellett (approved):

content/languages/riscv/index.md
```diff
@@ -10,9 +10,29 @@ Beta
...
+    - `Zbb` - Basic
+    - `Zbc` - Carry-less
+    - `Zbs` - Single-bit
+- Cryptography - Scalar<sup>§</sup>
+    - `Zbkx` - Crossbar permutation
+    - `Zk` - Standard
+    - `Zks` - ShangMi
+
+The most recent version of the specification for each can be downloaded below (as of 26 Dec 2023):
+- <sup>*</sup>[RISC-V Unprivileged Specification](https://github.com/riscv/riscv-isa-manual/releases/download/riscv-isa-release-056b6ff-2023-10-02/unpriv-isa-asciidoc.pdf)
```
error256, line 32: These aren't recognized as markdown links: https://deploy-preview-492--reverent-edison-2864ea.netlify.app/languages/riscv/

error256 (changes requested):

content/languages/riscv/index.md
StarsDown64: Fixed

error256 (approved):

PR #513, opened 2025-04-18
Adjusted link to point to wiki

Th0ught09:
The link on translation currently points to itself (and it doesn't have a lot of into), it confused me a lot when I wanted to try translations

The github wiki page is brilliant, I think it's a good idea to point there until the website has a better translation doc

### DIFF
--- a/content/authoring/translation.md
+++ b/content/authoring/translation.md
@@ -5,7 +5,7 @@ tags: [authoring]
 
 _TBD: improve this stub_
 
-New translators may refer to the [tutorial(TODO)](/authoring/translation/).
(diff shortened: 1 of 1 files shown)


hobovsky:
I completely forgot about the wiki howtos. Maybe the wiki page would be good candidate to be moved to the docs space? What do you think?

Th0ught09:
If you think it's a good idea I'm willing to do it :)

hobovsky:
If you are willing to spend some effort on that, then go ahead. Someone will definitely find this useful, and we will have a page which we can link when someone asks. It can be improved if/when needed.

PR #519, opened 2025-06-13
addition to the "Troubleshooting your solution" page

Summer-the-coder:
from my experience, `ERROR: UndefinedBehaviorSanitizer: SEGV on unknown address 0x000000000000` can also happen when an attempt to divide by zero is made, so I added it as a possible cause of this error.

### DIFF
--- a/content/training/troubleshooting.mdx
+++ b/content/training/troubleshooting.mdx
@@ -202,7 +202,9 @@ UndefinedBehaviorSanitizer can not provide additional info.
 ==1==ABORTING
 ```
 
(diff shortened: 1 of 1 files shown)


Summer-the-coder:
Yes (it's FPE instead of SEGV, but it's still "undefined behavior sanitizer"):

```cpp
#include <cmath>

bool isPrime(int n) {
  if (n < 2) { return false; }
  if (n < 4) { return true; }
  if (n % 2 == 0 || n % 3 == 0) { return false; }
  int stopAt = ((int) std::sqrt(n)) + 1;
  for (int i = 0; i < stopAt; i += 6) {
    if (n % i == 0 || n % (i + 2) == 0) {
      return false;
    }
  }
  
  return true;
}
```

Run this code in this kata: https://www.codewars.com/kata/5262119038c0985a5b00029f/cpp.

hobovsky:
I am trying to figure out a way to organize this section nicely, because errors causing `SEGV` are something different than `FPE` and I think they look strange placed both under a section which is titled as **`SEGV`**.

